PR #2, opened 2017-05-12
Add initial code for Puppet Language Server

glennsarti:
Previously the Puppet Extension only supported inprocess features.  This commit
adds an early version of a Language Server, which complies to the Language
Server protcol, which understands the Puppet Language.  Note this is not yet
ready to be packaged into the client.


jpogran:
Can the top level folder be called server and not `puppet-languageserver`?

jpogran:
How do we start this?

glennsarti:
@jpogran Reworked the PR as requested.

- now in `server` folder
- moved gem vendor to separate commit
- removed tests from vendored gem
- added brief readme for vendored gems and lang. server

PR #5, opened 2017-05-17
Add Node Graph Preview command

glennsarti:
Live node graph preview


glennsarti:
Example view
image

glennsarti:
Ready for merge.

jpogran:
Does a node graph only handle resources? Or should it handle classes?

jpogran:
Question answered offline. Only suggestion is to be able to choose side or steal focus of current window.

PR #14, opened 2017-05-18
Add better completion items - part 2

glennsarti:


jpogran:
Are these commits really maint? Is there maint in this project when its so new?

PR #16, opened 2017-05-18
Add rake tasks, packaging task and initial rspec tests

glennsarti:
This commit adds rubocop, rake and rspec for development tooling.  This commit
also adds a basic rubocop configuration file and an initial rubocop exemption
file.

Previously the Puppet Language Server initialisation was performed in
lib\puppet-languageserver.rb however this it unit test difficult as actions were
already taken.  This commit moves the initialisation to the root file which
does not get invoked during tests.

This commit adds the code for running unit and integration tests using rspec.
This commit also adds some trivial tests as a starting point.

Previously the language server source files were not transferred to the client
directory so they could be consumed by the client.  This commit adds a rake
task to copy the relevant files to the client.  This commit also adds a root
rakefile which can be used to package the extension into a vsix file.


Rakefile
```diff
@@ -0,0 +1,18 @@
...
+  puts "----------------------------"
+  puts "Building the Language Server"
+  puts "----------------------------"
+  Rake::Task["build"].invoke
+
+  puts "----------------------------"
+  puts "Building the Extension"
+  puts "----------------------------"
+  FileUtils.cd('client') do
+    `node ./node_modules/vsce/out/vsce package`
```
jpogran: Rakefile? When you only have a hammer...

Gemfile
```diff
@@ -0,0 +1,3 @@
+source ENV['GEM_SOURCE'] || "https://rubygems.org"
```
jpogran: Not sure I like ruby stuff leaking out of the server folder. While Puppet is ruby, vscode is not. I wonder if we should use the tooling the community would expect

client/.gitignore
```diff
@@ -0,0 +1,6 @@
+out/
+node_modules/
+vendor/
+
+// Ignore generated module output
+*.vsix
```
jpogran: This will interfere with putting up the vsix in the repo

PR #21, opened 2017-05-21
Add puppet resource support in Language Server

glennsarti:
Add puppet resource support in Language Server


client/src/providers/previewNodeGraphProvider.ts
```diff
@@ -21,17 +22,6 @@ export function getNodeGraphUri(uri: vscode.Uri) {
   });
 }
 
-
-export namespace CompileNodeGraphRequest {
```
jpogran: Nit, this commit in separate PR

PR #28, opened 2017-05-23
Update README with Language Server information

glennsarti:
The README had not been updated with the latest language server information.
This commit updates the readme and adds a nice animated gif showing the
extension in action.


jpogran:
Not sure this works with the vsgallery format, but will worry about that later

PR #31, opened 2017-05-24
Extension.ts cleanup

jpogran:
This cleans up extension.ts by moving out the custom Puppet Commands and Language server initialization to their own files.


glennsarti:
The new files appear to missing a newline at the end.  Apart from that LGTM!

PR #32, opened 2017-05-24
Add more default settings

jpogran:
This adds file encoding, file associations, and inserting a last new line on save to the default config".

PR #31 should be merged first


glennsarti:
Does this need a rebase or all good?

jpogran:
Rebased

Issue #30, opened 2017-05-24
Language server doesn't stop on client exit


jpogran:
```
--- INBOUND
{"jsonrpc":"2.0","id":2,"method":"shutdown","params":null}
---
[DEBUG] Received shutdown method
--- OUTBOUND
{"jsonrpc":"2.0","id":2,"result":null}
---
--- INBOUND
{"jsonrpc":"2.0","method":"exit","params":null}
---
[INFORMATION] Received exit notification.  Shutting down.
^C[TCPSRV] Started shutdown process. Press ^C to force quit.
[TCPSRV] Stopping services
[TCPSRV] Stopped listening on 127.0.0.1:8081
[TCPSRV] Waiting for workers to cycle down
[INFORMATION] Language Server exited.
```

I had to ctrl+c to actually exit

PR #33, opened 2017-05-25
Fix Language Server not stopping when client disconnects

glennsarti:
Previously when the client sent an `exit` message, the language server would
attempt to close the underlying connection however it would hang, and at worst
cause high CPU usage.  This commit
- Uses the callback architecture inside the TCP Server to close connection
  instead of trying to close the TCP Socket within the same execution context.
  This stops the possibility of a race condition which can cause unexpected
  blocks
- Add a missing method `close_connection_after_writing` on the base connection
  class
- Add more logging on TCP startup to describe whether the server will close
  after the client disconnects

### DIFF
--- a/server/lib/puppet-languageserver/json_rpc_handler.rb
+++ b/server/lib/puppet-languageserver/json_rpc_handler.rb
@@ -55,7 +55,7 @@ module PuppetLanguageServer
     end
 
     def unbind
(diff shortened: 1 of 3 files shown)


glennsarti:
Solves Issue #30

jpogran:
Works on My Machine&copy;!

PR #36, opened 2017-05-25
Speedup server startup

glennsarti:
NOTE - This builds on https://github.com/jpogran/puppet-vscode/pull/26

Previously the TCP Server would only start after parts of Puppet were
initialised.  This commit changes the startup behaviour so that the TCP Server
is running as soon as the Puppet Version can be determined, and further
initialisation is handed to a different thread.  A new command line option of
--start-slow is added, mainly for testing, to slow down the startup if required.

This will help when the server is started by the VSCode client directly so that
the TCP socket is available as soon as possible.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -76,6 +76,7 @@ Usage: puppet-languageserver.rb [options]
     -t, --timeout=TIMEOUT            Stop the language server if a client does not connection within TIMEOUT seconds.  A value of zero will not timeout.  Default is 10 seconds
     -d, --no-preload                 Do not preload Puppet information when the language server starts.  Default is to preload
         --debug=DEBUG                Output debug information.  Either specify a filename or 'STDOUT'.  Default is no debug output
+    -s, --slow-start                 Delay starting the TCP Server until Puppet initialisation has completed.  Default is to start fast
     -h, --help                       Prints this help
 ```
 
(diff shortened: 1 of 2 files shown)


glennsarti:
The biggest delay is in creating the puppet function envionment.  On my laptop it was 2 seconds or more.

Before
```
[INFORMATION] Using Puppet v4.10.1
[INFORMATION] Initializing settings...
[INFORMATION] Creating puppet function environment...
[INFORMATION] Using Facter v2.4.6
[INFORMATION] Preloading Facter (Async)...
[INFORMATION] Preloading Puppet Types (Async)...
[INFORMATION] Preloading Functions (Async)...
[INFORMATION] Starting RPC Server...
[TCPSRV] Services running. Press ^C to stop
[TCPSRV] Will stop the server in 2 seconds if no connection is made.
[TCPSRV] Will stop the server when client disconnects
[TCPSRV] Started listening on 127.0.0.1:8081.
[TCPSRV] No connection has been received in 2 seconds.  Shutting down server.
...
```

Now with fast-start
```
[INFORMATION] Using Puppet v4.10.1
[INFORMATION] Starting RPC Server...             <---------- MUCH EARLIER START
[TCPSRV] Services running. Press ^C to stop
[TCPSRV] Will stop the server in 2 seconds if no connection is made.
[TCPSRV] Will stop the server when client disconnects
[INFORMATION] Initializing settings...
[TCPSRV] Started listening on 127.0.0.1:8081.
[INFORMATION] Creating puppet function environment...
[INFORMATION] Using Facter v2.4.6
[INFORMATION] Preloading Facter (Async)...
[INFORMATION] Preloading Puppet Types (Async)...
[INFORMATION] Preloading Functions (Async)...
...
```

glennsarti:
Rebased.

Issue #54, opened 2017-05-30
`Puppet: Resource` command does not match the output from 'puppet resource ..'


glennsarti:
doing a Puppet: Resource command generates the following for a group:
```puppet
group { 'Administrators':
  ensure               => 'present',
  attribute_membership => 'minimum',
  gid                  => 'S-1-5-32-544',
  loglevel             => 'notice',
  members              => ['Administrator', 'glenn.sarti'],
  provider             => 'windows_adsi',
}
```

This does not match the output from `puppet resource group`

I suspect the language server is pulling the puppet resource information too early in the code path

glennsarti:
Fixed in PR #55

glennsarti:
Resolved

PR #55, opened 2017-05-31
(GH-54) Fix Puppet Resource Command

glennsarti:
Previously the output from the Puppet Resource command in the language server
did not match the output from a normal `puppet resource` invocation.  This
commit modifies the language server to mimic the behaviour from puppet-agent
by use the prune_parameters method prior to send the data back to the client.
Also and additional newline is placed between each resource for better
readability.


glennsarti:
Fixes #54

Issue #7, opened 2017-05-18
Integrate Puppet Language-Server into Extension


jpogran:
Currently you have to start the ruby language server manually before running the extension either in debug or in production mode. Ideally we would have the language server start as the extension is loaded

jpogran:
PR #56

jpogran:
This is now resolved

PR #56, opened 2017-05-31
(GH-7) Start lang server without wait

jpogran:
This PR starts the ruby language server without using a delay or sleep when the extension starts, so there is no need to start it yourself

- [x] Start ruby language server async
- [x] Find ruby language server path dynamically
- [ ] Find where Puppet is dynamically
- [ ] Config file setting for puppet path


client/src/languageserver.ts
```diff
@@ -3,11 +3,54 @@ import {
...
+  var serverPath = path.join(extensionPath, 'server');
+  // "/some/file/structure/vscode-puppet/server/puppet-languageserver"
+  var serverExe = path.join(serverPath, 'puppet-languageserver');
+
+  let cmd = '';
+  let args = [serverExe];
+  let options = {};
+
+  if (process.platform == 'win32') {
+    cmd = 'c:\\windows\\system32\\cmd.exe'
```
glennsarti: Need to change this to sysnative on 64bit.  Vscode is a 32bit processs on Windows.  Should probably use %COMSPEC%  too?

jpogran:
This leaves off windows for the moment until we can figure out process spawn correctly there. This also leaves off packaging, as it's now not working again. Need to research more, but both these issues are out of scope for this PR

client/src/extension.ts
```diff
@@ -1,30 +1,61 @@
...
 
-  createStatusBarItem();
+    let config = vscode.workspace.getConfiguration('puppet');
 
-  var languageServerClient = startLangServerTCP(host, port, langID, statusBarItem);
-  context.subscriptions.push(languageServerClient.start());
+    var host             = config['languageserver']['address']; // '127.0.0.1';
+    var port             = config['languageserver']['port']; // 8081;
+    var stopOnClientExit = config['languageserver']['stopOnClientExit']; // true;
+    var timeout          = config['languageserver']['timeout']; // 8081;
```
glennsarti: nit: It's not 8081 for the timeout

client/src/languageserver.ts
```diff
@@ -1,14 +1,53 @@
...
 import * as cp from 'child_process';
 import ChildProcess = cp.ChildProcess;
+import {
+  LanguageClient, LanguageClientOptions, ServerOptions
+} from 'vscode-languageclient'
 
 import * as messages from '../src/messages';
 
-export function startLangServerTCP(host: string, port: number, langID: string, statusBarItem): LanguageClient {
+export function createLanguageServerProcess(serverExe: string, myOutputChannel: vscode.OutputChannel) {
```
glennsarti: nit: should it be called Exe?  that's more a Windows-ism

jpogran: what should it be called? binary? file?

client/src/extension.ts
jpogran: copy-pasta comments

Issue #12, opened 2017-05-18
Add CI PR workflow for AppVeyor (windows) and Travis (*nix)


glennsarti:

glennsarti:
Should remember to test on different puppet versions.  Perhaps 4.7 (LTS) and latest?

PR #71, opened 2017-06-09
(GH-12) Add Travis file

jpogran:
Add travis file, using https://code.visualstudio.com/docs/extensions/testing-extensions#_running-tests-automatically-on-travis-ci-build-machines as example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+language: node_js
+node_js:
+  - "node"
+
+environment:
+  ELECTRON_RUN_AS_NODE: 1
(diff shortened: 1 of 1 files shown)


glennsarti:
In Appveyor you may want to tick the "Ignore branches with no appveyor" checkbox.

glennsarti:
`cd client` is in there twice

jpogran:
Yes, was unsure of state. However, that's not what failing

jpogran:
This is fixed, had to use https://github.com/Microsoft/vscode/blob/master/.travis.yml as an example

Issue #49, opened 2017-05-30
Completion doesn't work in a resource in a new line


glennsarti:
Completion doesn't work in a resource in a new line

Doesn't resolve params/properties at X
```puppet
user { 'Bob':
  ensure => present,
X
}
```

DOES resolve params/properties at X
```puppet
user { 'Bob':
  ensure => present,

X
}
```

Need to change the document parser to  not always go back one character if it's already at the beginning of a line.

glennsarti:
Resolved in https://github.com/jpogran/puppet-vscode/pull/72

PR #72, opened 2017-06-09
Fix completion at file beginning on new lines and on keywords

austb:
I wanted to get a PR open to get some comments on it.

Changes

1. remove_char replaced by remove_word this allows users to get autocomplete if they start typing on a partial word such as "ens". Previously, they only received completion suggestions when typing a new word. remove_chars is left at the end of the list to facilitate completion of `$facts[`.
1. removed the `if abs_offset > 0` because it was preventing the start of the document from being interpreted as the root of a document, which meant no completion was being generated.
1. For completion only, added a check that the item isn't a `Puppet::Pops::Model::QualifiedName`, this always seems to be the "deepest" part of the AST if the user is typing a word, but it prevents useful autocompletion from being generated when typing at the root of the manifest.

### DIFF
--- a/server/lib/puppet-languageserver/puppet_parser_helper.rb
+++ b/server/lib/puppet-languageserver/puppet_parser_helper.rb
@@ -1,13 +1,26 @@
 module PuppetLanguageServer
   module PuppetParserHelper
-    def self.remove_char_at(content, line_offsets, line_num, char_num)
+    def self.remove_chars_starting_at(content, line_offsets, line_num, char_num, num_chars_to_remove)
       line_offset = line_offsets[line_num]
       raise if line_offset.nil?
 
       # Remove the offending character
-      new_content = content.slice(0, line_offset + char_num - 1) + content.slice(line_offset + char_num, content.length - 1)
+      new_content = content.slice(0, line_offset + char_num - num_chars_to_remove) + content.slice(line_offset + char_num, content.length - num_chars_to_remove)
 
       new_content
+    end
+
+    def self.remove_char_at(content, line_offsets, line_num, char_num)
+      remove_chars_starting_at(content, line_offsets, line_num, char_num, 1)
+    end
+
+    def self.get_char_at(content, line_offsets, line_num, char_num)
+      line_offset = line_offsets[line_num]
+      raise if line_offset.nil?
+
+      absolute_offset = line_offset + (char_num - 1)
+
+      content[absolute_offset]
     end
 
     def self.insert_text_at(content, line_offsets, line_num, char_num, text)
@@ -43,7 +56,7 @@ module PuppetLanguageServer
       end
     end
 
-    def self.object_under_cursor(content, line_num, char_num, multiple_attempts = false)
+    def self.object_under_cursor(content, line_num, char_num, multiple_attempts = false, disallowed_classes = [])
       # Use Puppet to generate the AST
       parser = Puppet::Pops::Parser::Parser.new
 
@@ -53,7 +66,7 @@ module PuppetLanguageServer
 
       result = nil
       move_offset = 0
-      %i[noop remove_char try_quotes try_quotes_and_comma].each do |method|
+      %i[noop remove_word try_quotes try_quotes_and_comma remove_char].each do |method|
         new_content = nil
         case method
         when :noop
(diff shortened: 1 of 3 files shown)


server/lib/puppet-languageserver/puppet_parser_helper.rb
```diff
@@ -53,14 +66,23 @@ def self.object_under_cursor(content, line_num, char_num, multiple_attempts = fa
...
         case method
         when :noop
           new_content = content
         when :remove_char
           new_content = remove_char_at(content, line_offsets, line_num, char_num)
           move_offset = -1
+        when :remove_word
+          next_char = get_char_at(content, line_offsets, line_num, char_num)
+
+          while /\w/.match(next_char)
```
glennsarti: need to watch out for this.  regex `\w` doesn't support UTF8 chars. We had similar issues in puppet codebase

austb: I'll switch it to `/[[:word:]]/` which will match ["a character in one of the following Unicode general categories Letter, Mark, Number, Connector_Punctuation"](http://ruby-doc.org/core-2.3.1/Regexp.html)

austb:
@jpogran I changed this milestone back to 0.4.0 because it is for Issue #49 which is a part of that milestone. Let me know if that's incorrect.

server/lib/puppet-languageserver/puppet_parser_helper.rb
```diff
@@ -1,15 +1,28 @@
 module PuppetLanguageServer
   module PuppetParserHelper
-    def self.remove_char_at(content, line_offsets, line_num, char_num)
+    def self.remove_chars_starting_at(content, line_offsets, line_num, char_num, move_offset)
```
glennsarti: Is `move_offset` the best name for the parameter here?  Is this the number of chars to remove?

server/lib/puppet-languageserver/puppet_parser_helper.rb
```diff
@@ -1,15 +1,28 @@
...
       line_offset = line_offsets[line_num]
       raise if line_offset.nil?
 
       # Remove the offending character
-      new_content = content.slice(0, line_offset + char_num - 1) + content.slice(line_offset + char_num, content.length - 1)
+      new_content = content.slice(0, line_offset + char_num + move_offset) + content.slice(line_offset + char_num, content.length - 1)
```
glennsarti: Need to modify both slice commands I believe...

server/lib/puppet-languageserver/puppet_parser_helper.rb
```diff
@@ -1,15 +1,28 @@
...
 
+    def self.remove_char_at(content, line_offsets, line_num, char_num)
+      remove_chars_starting_at(content, line_offsets, line_num, char_num, -1)
+    end
+
+    def self.get_char_at(content, line_offsets, line_num, char_num)
+      line_offset = line_offsets[line_num]
+      raise if line_offset.nil?
+
+      absolute_offset = line_offset + (char_num - 1)
```
glennsarti: Internally the `line_num` and `char_num` are zero based (Puppet uses 1 based) so I don't see why you're using `char_num - 1` here.

server/lib/puppet-languageserver/puppet_parser_helper.rb
```diff
@@ -53,14 +66,23 @@ def self.object_under_cursor(content, line_num, char_num, multiple_attempts = fa
...
         case method
         when :noop
           new_content = content
         when :remove_char
           new_content = remove_char_at(content, line_offsets, line_num, char_num)
           move_offset = -1
+        when :remove_word
+          next_char = get_char_at(content, line_offsets, line_num, char_num)
+
+          while /[[:word:]]/.match(next_char)
```
glennsarti: This loop looks like it could use a closing terminator if move_offset gets too big

server/lib/puppet-languageserver/puppet_parser_helper.rb
austb: When typing the "n" in this manifest, I ran into the following problem.
```
user { 'Bob' :
  ensure => 'present',
  n
}
```

Subtracting 1 got me the character I wanted, but maybe I should have gone back one character in the `remove_word` method instead?
```
    18: def self.get_char_at(content, line_offsets, line_num, char_num)
    19:   line_offset = line_offsets[line_num]
    20:   raise if line_offset.nil?
    21:
    22:   binding.pry
    23:   absolute_offset = line_offset + (char_num - 1)
    24:
 => 25:   content[absolute_offset]
    26: end

[1] pry(PuppetLanguageServer::PuppetParserHelper)> content[absolute_offset]
=> "n"
[2] pry(PuppetLanguageServer::PuppetParserHelper)> content[absolute_offset + 1]
=> "\n"
```

server/lib/puppet-languageserver/puppet_parser_helper.rb
austb: What would a reasonable termination length be?

server/lib/puppet-languageserver/puppet_parser_helper.rb
glennsarti: Ahh yes,....that problem.

absolute_offset of zero is the end of the first character i.e. the cursor is on the right hand side of the specified location.  Thus the `-1`.

server/lib/puppet-languageserver/puppet_parser_helper.rb
glennsarti: It should not exceed `char_num` Otherwise we've moved backwards to the previous line

glennsarti:
Just need to squash the commits and we're good for merge.

PR #76, opened 2017-06-15
{WIP} Rubocop fixes

austb:
**Changes**

- I do not think that the Rubocop rule to separate long numbers with underscores as commas is a useful thing. I disabled that feature.

**Questions**

- `Style/StringLiterals` Should we disabled this? I don't think it's worthwhile to enforce `'` vs `"` usage.

**Remaining Issues**

- [x] Global variable used for logging
- [ ] "Rescue all exceptions" rubocop would like these to be more specific


austb:
Github is being dumb and not updating this PR with new commits. Closing for a new one.

PR #77, opened 2017-06-16
Fix a bunch of rubocop errors

austb:


glennsarti:
Manually test on Windows.  All seems to work.

Issue #65, opened 2017-06-07
Puppet Language Server Windows support


jpogran:
As of PR #7, the language server starts with the extension automatically on all supported OS, with the exception of windows. We had trouble starting the language server and keep it running while the extension is loaded using node and child_process.

We should support windows the same as the other platforms.

glennsarti:
@jpogran You happy to close this off?

jpogran:
@glennsarti will test on a fresh VM to avoid the 'works on my machine' problem we both have, then we will close this

jpogran:
After further review, this can be closed as it's tested

glennsarti:
@jpogran Verified it works as expected on Server 2012R2 64bit and Windows 10 32bit.  Required PR #85 first though to create an installable extension.

PR #79, opened 2017-06-20
(GH-65) Start language server on Windows

glennsarti:
This commit adds Windows support to the local Language Server. It detects when
a 32bit VS Code is on a 64bit OS and will use the appropriate file system
redirection to find a Puppet Agent installation.  This commit also improves the
user experience and will not throw large errors, but instead warns the user that
the language server could not be started and has reduced functionality.

This commit updates the local language server invocation to honor the extension
settings, for example, port or debug file.  This commit removes the
stopOnClientExit property as it does not make sense on the local language server
and will cause issues on extension restart.  This commit adds a debugFilePath
setting for the local language server to direct all debug output to this file.


glennsarti:
Builds on PR #80

PR #81, opened 2017-06-20
Add better puppet-agent detection

glennsarti:
Previously the default behaviour was to use ruby in the path however some puppet
agent installations do not put its ruby on the path and is instead located in
/opt/puppetlabs/puppet/bin/ruby. This commit adds file detection to prefer the
specific puppet agent location of ruby and fall back to just the path.


glennsarti:
Re-targetted to master

Issue #98, opened 2017-07-07
Language server does not return when creating function environment


glennsarti:
When starting the PLS get the following;

```
C:\pls\puppet-vscode\server>ruby puppet-languageserver --timeout=0 --ip=0.0.0.0
--no-stop --debug=STDOUT
I, [2017-07-07T22:41:53.716708 #2800]  INFO -- : Language Server is v0.0.1
I, [2017-07-07T22:41:53.716708 #2800]  INFO -- : Using Puppet v4.10.1
I, [2017-07-07T22:41:53.716708 #2800]  INFO -- : Initializing settings...
I, [2017-07-07T22:41:53.716708 #2800]  INFO -- : Starting RPC Server...
D, [2017-07-07T22:41:53.732306 #2800] DEBUG -- : TCPSRV: Services running. Press
 ^C to stop
LANGUAGE SERVER RUNNING 0.0.0.0:8081
D, [2017-07-07T22:41:53.732306 #2800] DEBUG -- : TCPSRV: Started listening on 0.
0.0.0:8081.
I, [2017-07-07T22:41:53.747929 #2800]  INFO -- : Creating puppet function enviro
nment...
```
It never returns

Env:
Name - jenkins-slave-windows2012-dev-1
OS - Server 2012R2
Puppet - 4.10.1 (using ruby from there)

This is a managed node that has many classes etc. in the cache directory.  Could be a circular reference in there somewhere.

glennsarti:
Merged and be closed.

PR #101, opened 2017-07-13
(GH-98) Fix function and type loading initial fix

glennsarti:
Previously type and function loading would prematurely fail and stop the
language server if a type or function load threw an error.  This appears to be
a facet of the Puppet loadall function.  This PRchanges the loading to
mimic the loadall function but instead, ignore any loading errors and continue.
This PR also adds some debug messages after types and function have
completed loading which is useful during debugging.

### DIFF
--- a/server/lib/puppet-languageserver.rb
+++ b/server/lib/puppet-languageserver.rb
@@ -118,11 +118,7 @@ module PuppetLanguageServer
   end
 
   def self.init_puppet_worker(options)
-    Puppet.initialize_settings
-
-    log_message(:info, 'Creating puppet function environment...')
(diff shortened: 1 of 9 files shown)


glennsarti:
I can't say I've used Monitors before.  After a brief glace;
* I don't want the code to be re-entrantable as I'm protecting against accessing an instance variable
* There is a performance hit for using Monitors over Mutexes.  Not sure it will be observable in our use case though.
* I kind of like the Mutexes erroring when it detects a deadlock.  While not applicable to us as we will only ever be using MRI ruby, but on JRuby where we have real threading, deadlocking is bad.

PR #111, opened 2017-08-22
(GH-103) Parse puppet-lint.rc in module directory

### LINKED ISSUE #103: puppet-lint.rc is not honored
## Summary

Currently, the extension's bundled version of puppet-lint (maybe bundled is the wrong word?) does not read or use a user's `puppet-lint.rc` configuration file.

https://github.com/rodjek/puppet-lint#configuration-file

This configuration file is used to disable puppet-lint checks. A common one being disabling the 140 character check.

## Expected behavior

Puppet-lint checks that I've disabled in my `~/.puppet-lint.rc` file don't show up in VSCode.

## Actual behavior

VSCode shows all puppet-lint checks even though I've disabled some of them in my `~/.puppet-lint.rc` configuration file.

## Preferred Solution

`puppet-lint.rc` is parsed and consumed by the extension to disable checks in the order shown here (pulled from the [puppet-lint README](https://github.com/rodjek/puppet-lint#configuration-file)): 

1. /etc/puppet-lint.rc
1. ~/.puppet-lint.rc
1. .puppet-lint.rc

## Alternate Solution

The extension could have workspace settings to disable arbitrary puppet-lint checks:

```json
{
	"puppet.languageserver.disabled_lint_checks": [
		"140-chars-check",
		"documentation-check"
	]
}
```

austb:
This adds the ability for a module-level .puppet-lint.rc file to be
honored by the language server, when the language server is started
locally with --local-workspace=PATH command

Fixes #103.

### DIFF
--- a/client/src/connection.ts
+++ b/client/src/connection.ts
@@ -224,6 +224,7 @@ export class ConnectionManager implements IConnectionManager {
 
     if ((this.connectionConfiguration.host == undefined) || (this.connectionConfiguration.host == '')) {
       args.push('--ip=127.0.0.1');
+      args.push('--local-workspace=' + vscode.workspace.textDocuments[0]);
     } else {
       args.push('--ip=' + this.connectionConfiguration.host);
     }
--- a/server/lib/puppet-languageserver.rb
+++ b/server/lib/puppet-languageserver.rb
@@ -24,7 +24,8 @@ module PuppetLanguageServer
         connection_timeout: 10,
         preload_puppet: true,
         debug: nil,
-        fast_start_tcpserver: true
+        fast_start_tcpserver: true,
+        workspace: nil,
       }
 
       opt_parser = OptionParser.new do |opts|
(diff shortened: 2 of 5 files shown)


client/src/connection.ts
```diff
@@ -224,6 +224,7 @@ export class ConnectionManager implements IConnectionManager {
 
     if ((this.connectionConfiguration.host == undefined) || (this.connectionConfiguration.host == '')) {
       args.push('--ip=127.0.0.1');
+      args.push('--local-workspace=' + vscode.workspace.textDocuments[0]);
```
austb: Could not tell if this was the best way to do it. There's also [workspaceFolders](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_workspace).

jpogran: Theres `cwd`, but the danger of these is that they are null if you don't open a folder with vscode

austb: Ya, that's why I chose textDocuments. It looked like it'd be populated with all the files of a workspace. So just pick one and work from there to metadata.json

bastelfreak:
Hey people,
I'm a bit lost here. My vscode with puppet-vscode 1.4.0 plugin ignores the .puppet-lint.rc in a module root directory. Based on this PR, I'm not sure if I need to configure anything? And I think I don't need to explicitly start a languageserver as that's bundles with the vscode plugin?

bastelfreak:
Is it possible that I need to open each module as workspace? I've my control repo cloned + all modules into that and the control repository is my workspace in vscode.

PR #112, opened 2017-08-22
(GH-109) Randomize language server port when local

### LINKED ISSUE #109: Randomize language server port
See PHP extension for example

austb:
When the language server starts locally, it will connect on a random
port to allow multiple workspaces to be open at one time. When the
connection is not identified as a local connection, it will still use
the specified settings or default.

Fixes #109.


jpogran:
Refactoring to methods and changing it to random ports makes it hard to figure out whats going on.

austb:
From what Glenn said, you could not work on two local workspaces at once without random ports.

jpogran:
Rebased and fixed conflict

PR #114, opened 2017-08-24
adds basic PDK support

jpogran:
This adds basic support for the new PDK from Puppet

~~- [ ]  Detect platform and switch path to PDK for new module~~
- [x]  Use existing shell for pdk commands
- [x]  Add telemetry for pdk commands

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -5,3 +5,4 @@ tmp/
 
 // Ignore generated module output
 *.vsix
+.vscode-test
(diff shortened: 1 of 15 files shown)


jpogran:
This takes the approach of relying on PDK being installed and on path (mac/linux) or module loaded (windows).

It will open a new window to create a module, but use the existing window to create a class or run validate/test commands. It uses the terminal for every command except new module

PR #117, opened 2017-09-07
Improved command titles

turbodog:
Improving clarity of command titles:
- Explicitly show CLI commands, e.g. pdk new class, plus a description of what it does
- Have "Puppet" in all titles so that typing Puppet shows all available commands

Note: I don't know if using this style is conventional for VS Code or plugins. Feedback on that desired.

### DIFF
--- a/client/CHANGELOG.md
+++ b/client/CHANGELOG.md
@@ -6,11 +6,13 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## Unreleased
 
-## 0.6.1
-
+* Added support for the PDK(https://puppet.com/blog/develop-modules-faster-new-puppet-development-kit)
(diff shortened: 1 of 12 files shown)


jpogran:
This is how things look now. Notice that `Puppet:` is always prepended to the text, this shows which extension the command is from. This already satisfies item 2.

current_command_titles

With your suggested changes, it looks a little repetitious to me:

changed_command_titles

I understand the reason for putting the actuall command cased liked it would be in the command line, to try to educate the user on what command to type themselves, but FWIW the common style is to used descriptive sentences and not actual commands. Some of the ones below are commands that run exes, but details of that are not shown:

normal_command_titles

I think the output in the terminal after every command execution satisfies the need to show the command:

terminal_output

I'll play around with a couple more approaches

jpogran:
@turbodog let me know if the new command titles are fine by you. I'm closing out this PR for now as it's out of date and should be resolved by the recent 0.7.1 release

turbodog:
LGTM

Issue #130, opened 2017-09-15
Use the PDK ruby instance if the native puppet-agent does not exist


glennsarti:
Some people are unable to install a native puppet-agent, but are able to instead install the PDK.  The PDK does include a native ruby environment as well as puppet gem and its required gems.  This means the VS Code language server will be able to take advantage of this.

This ticket is to add detection of a PDK setup and use that ruby environment if a native agent is not available.  As this is a cross platform tools, this detection should also be cross platform

glennsarti:
Removing from 0.7 milestone as the prep is done but not enabled yet for 0.7

glennsarti:
@jpogran This can probably be closed now?

jpogran:
@glennsarti Yep

PR #134, opened 2017-09-16
(GH-130) Prepare to use PDK ruby if puppet agent not available

glennsarti:
Previously the language server required a puppet-agent installation however
some users are not able to install a puppet-agent, but instead the PDK.  This
commit changes the connect.ts script;

- Refactors the ruby detection into two helpers, getLanguageServerFromPDK and
  getLanguageServerFromPuppetAgent.  These helpers are responsible for detecting
  and setting up the process creation for their respective ruby environments on
  any platform type
- Removes the need for environment.bat on Windows by setting up the process env
  variables at process creation
- Updates the logging to be more verbose in the detection methods

Note the functionality to actually use the PDK is currently disabled as the
configuration settings and user story need to be completed first.  However to
not lose the effort, the code still exists but is commented out.

### DIFF
--- a/client/src/connection.ts
+++ b/client/src/connection.ts
@@ -209,12 +209,31 @@ export class ConnectionManager implements IConnectionManager {
     callback(proc);
   }
 
-  private createLanguageServerProcess(serverExe: string, callback : Function) {
-    this.logger.debug('Language server found at: ' + serverExe)
+  private getDirectories(parent) {
+    return fs.readdirSync(parent).filter(function (file) {
+      return fs.statSync(path.join(parent, file)).isDirectory();
+    });
+  }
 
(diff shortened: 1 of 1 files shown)


glennsarti:
Note this PR does not add support for PDK on non-Windows.  Need to do that still.  @jpogran Should we hold up this PR till then or not?

glennsarti:
Confirmed this works on Windows and Ubuntu 16.04 desktop edition.  Can be evaluated by:
- Setting `"puppet.languageclient.minimumUserLogLevel":"debug"` in user settings in VS Code
-  Setting `"puppet.puppetAgentDir": "somedirthatdoesnotexist"` in user settings to stop the Puppet Agent being detected and instead try and detect via PDK.

Issue #136, opened 2017-09-18
Create a better UI experience while puppet loads


glennsarti:
While the puppet language server loads the puppet types and functions, any requests for hover/autocomplete are held up via mutexes.

This is great as it marshals access to this information however the user does not know.  Instead we should display something to the user that shows the language server is running, but still not completely available.

Issue #61, opened 2017-06-05
Create a better experience when the Lang client fails to connect to the server


glennsarti:
At the moment there is no UI to indicate failure or reason for failure if the lang client fails to connect.  We should instead take some inspiration from the PowerShell extension and how it gracefully degrades and informs the user.

PR #137, opened 2017-09-18
(GH-136)(GH-61) Make client langserver better

glennsarti:
Note - This builds on https://github.com/jpogran/puppet-vscode/pull/135

Previously the Language client had no way to know whether the langauge server
had completed loading of facts, types and functions.  This caused requests to
be blocked because of mutex locks while this information is being loaded.

This commit;
- Adds three properties which the client can use to determine the loading status
  of various parts of the Language Server, with unit tests
- The puppet and facter helpers were modified to expose a *_loaded? method
- The puppet and facter helpers will now return empty datasets if they in the
  process of being loaded on another thread.  This speeds up responses to the
  client

The client will now query the language server periodically after connecting.
This query is for the load status which is shown in the UI of the extension
status bar item, until it is fully loaded.  Once loaded, the puppet version is
displayed.

Previously there was no way to restart the Language Client or view the Puppet
Client logs easily.  This commit adds a click menu which displays an option to
either restart the Lanugage Client or view the connection logs.  The location
to click is on the Language Server status bar item in the bottom right corner.

This commit changes the configuration in package.json to show the logging levels
are an enum, not free text, and adds timestamps to the log messages.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -38,6 +38,8 @@
   ],
   "activationEvents": [
     "onLanguage:puppet",
+    "onCommand:extension.puppetRestartSession",
+    "onCommand:extension.puppetShowConnectionLogs",
     "onCommand:extension.puppetShowConnectionMenu",
     "onCommand:extension.puppetLint",
     "onCommand:extension.puppetParserValidate",
@@ -139,6 +141,12 @@
     ],
     "menus": {
       "commandPalette": [
+        {
+          "command": "extension.puppetRestartSession"
+        },
+        {
+          "command": "extension.puppetShowConnectionLogs"
+        },
         {
           "command": "extension.pdkNewModule"
         },
--- a/client/package.json
+++ b/client/package.json
@@ -38,6 +38,7 @@
   ],
   "activationEvents": [
     "onLanguage:puppet",
+    "onCommand:extension.puppetShowConnectionMenu",
     "onCommand:extension.puppetLint",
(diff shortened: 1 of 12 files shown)


glennsarti:
The loading status looks like;
image

client/package.json
```diff
@@ -88,6 +89,11 @@
     ],
     "commands": [
       {
+        "command": "extension.puppetShowConnectionMenu",
+        "category": "Puppet",
+        "title": "Show Connection Menu"
```
glennsarti: Should we have the command show the menu? or should we just show the individual commands and the context menu just shows a curated list?

glennsarti:
Updated both the commands to be on the command palette:
image

glennsarti:
@jpogran Rebased onto master.

client/package.json
```diff
@@ -38,6 +38,7 @@
   ],
   "activationEvents": [
     "onLanguage:puppet",
+    "onCommand:extension.puppetShowConnectionMenu",
```
jpogran: This works without listing the two commands here?

client/package.json
```diff
@@ -88,6 +89,16 @@
     ],
     "commands": [
       {
+        "command": "extension.puppetRestartSession",
```
jpogran: These need to also be added to the command palette list. Lots of duplication in package.json

Issue #138, opened 2017-09-18
Update telemetry to prod


jpogran:
Update telemetry:

- update key to point to prod
- update commands to put command name directly in response

glennsarti:
Merged

PR #139, opened 2017-09-18
(GH-138) Update telemetry to prod

jpogran:


client/src/connection.ts
```diff
@@ -326,6 +327,16 @@ export class ConnectionManager implements IConnectionManager {
       langClient.logger.debug('Language server client started, setting puppet version')
       languageServerClient.sendRequest(messages.PuppetVersionRequest.type).then((versionDetails) => {
         this.setConnectionStatus(versionDetails.puppetVersion, ConnectionStatus.Running);
+        if (reporter) {
+          reporter.sendTelemetryEvent('puppetVersion' +versionDetails.puppetVersion);
+          reporter.sendTelemetryEvent('facterVersion' + versionDetails.facterVersion);
+          reporter.sendTelemetryEvent('languageServerVersion' + versionDetails.languageServerVersion);
+          reporter.sendTelemetryEvent('version', {
```
glennsarti: Curious as to why this is listed twice

jpogran: I'm unsure what format we want going forward, so until then log twice.

Issue #116, opened 2017-09-07
Publishing extension does not include languageserver


jpogran:

glennsarti:
Removing from milestone as this is not a release-able item.

PR #149, opened 2017-09-21
(GH-116) Fix packaging to conform with vsce workflow

glennsarti:
Previously the gulp file could be used to package the extension however that is
not how the VS Code Extension tool workflow operates (vsce).  This commit
changes the tool chain so that vsce calls the gulpfile during pre-publish to
ensure the language server is updated prior to packaging.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 0.0.4-appv.{build}
+version: 0.7.0-appv.{build}
 clone_depth: 10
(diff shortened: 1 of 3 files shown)


glennsarti:
Example artifact is here -> https://ci.appveyor.com/project/jpogran/puppet-vscode/build/0.7.0-appv.276/job/p3w68q1h3ww7nwtr/artifacts

Issue #157, opened 2017-09-29
Puppet Resource command hidden


jpogran:
Due to the way we set up the command palette for `puppet resource`, it hides it from being available. Fix the when clause to allow it to be activated

PR #158, opened 2017-09-29
(GH-157) Activate Puppet Resource Command

jpogran:
Due to the when clause used in the command palette setup for the  command, it was hidden from view to the user but available to be invoked. This commit changes visibility so now its viewable to the user


glennsarti:
Confirmed it's fixed on Windows.

Issue #97, opened 2017-07-07
Language server crashing


glennsarti:
The language server was crashing with:
```
D, [2017-07-07T22:25:01.996734 #3900] DEBUG -- : TCPSRV: Closed socket due to er
ror - undefined method `line' for #<ArgumentError:0x000000085c88e0>
["C:/pls/puppet-vscode/server/lib/puppet-languageserver/document_validator.rb:60
:in `rescue in block in validate'", "C:/pls/puppet-vscode/server/lib/puppet-lang
uageserver/document_validator.rb:49:in `block in validate'", "C:/Program Files/P
uppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby/puppet/context.rb:65:in `overrid
e'", "C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby/puppet.r
b:294:in `override'", "C:/pls/puppet-vscode/server/lib/puppet-languageserver/doc
ument_validator.rb:48:in `validate'", "C:/pls/puppet-vscode/server/lib/puppet-la
nguageserver/message_router.rb:132:in `receive_notification'", "C:/pls/puppet-vs
code/server/lib/puppet-languageserver/json_rpc_handler.rb:176:in `process'", "C:
/pls/puppet-vscode/server/lib/puppet-languageserver/json_rpc_handler.rb:131:in `
received_parsed_object'", "C:/pls/puppet-vscode/server/lib/puppet-languageserver
/json_rpc_handler.rb:123:in `parse_data'", "C:/pls/puppet-vscode/server/lib/pupp
et-languageserver/json_rpc_handler.rb:108:in `receive_data'", "C:/pls/puppet-vsc
ode/server/lib/puppet-languageserver/simple_tcp_server.rb:82:in `get_data'", "C:
/pls/puppet-vscode/server/lib/puppet-languageserver/simple_tcp_server.rb:206:in
`call'", "C:/pls/puppet-vscode/server/lib/puppet-languageserver/simple_tcp_serve
r.rb:206:in `fire_event'", "C:/pls/puppet-vscode/server/lib/puppet-languageserve
r/simple_tcp_server.rb:108:in `block in start'", "C:/pls/puppet-vscode/server/li
b/puppet-languageserver/simple_tcp_server.rb:110:in `call'", "C:/pls/puppet-vsco
de/server/lib/puppet-languageserver/simple_tcp_server.rb:110:in `block (2 levels
) in start'"]
```

[...]

PR #162, opened 2017-10-03
(GH-97) Create a crash file when the language server abends

glennsarti:
Previously when the language server crashed, no debug information would be
created which would make it difficult to diagnose issues in the field.  This
commit adds the facility to create a crash file in
%TMP%\puppet_language_server_crash.txt.

This commit also adds tests for this facility. #97


glennsarti:
@jpogran Ready for review/merge

PR #173, opened 2017-10-24
(docs)(maint) Replace all docs.puppet.com URLs

jtappa:
with new corresponding puppet.com/docs URLs.
During puppetconf this year, the current versions of docs were moved from their own URL (docs.puppet.com) to puppet.com/docs.

I tried to test out each replacement as I updated, they all appear to be working. There are of course redirects in place for these so I don't think any are currently 404ing as-is, but it's preferred we update the links and references themselves where possible.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -11,9 +11,9 @@ This extension provides full Puppet Language support for [Visual Studio Code](ht
 
 You will need to have Puppet Agent installed in order to fully use this extension. You can find instructions and installation links here:
 
-* [Windows](https://docs.puppet.com/puppet/latest/install_windows.html)
-* [MacOSX](https://docs.puppet.com/puppet/latest/install_osx.html)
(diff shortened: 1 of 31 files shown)


glennsarti:
Hi @jtappa .  Was this also in a particular version of puppet-lint?  As in this case we're just copying a specific version of the gem.

jpogran:
Thanks for the edits @jtappa, @glennsarti updated the vendored puppet-lint files. Can you rebase and push to your branch?

jtappa:
I didn't realize these were being populated by puppet lint, would it be better for me to open a PR there, so these don't get overwritten the next time you guys update?

glennsarti:
@jtappa 100% YES PLEASE!

glennsarti:
Closing this PR as it should be sent upstream.

Issue #180, opened 2017-10-27
Backslashes in File Path Do Not Display in Node Graph


michaeltlombardi:
### What Versions are you running?

+ **OS Version:** Windows 10
+ **VSCode Version:** 1.17.2
+ **Puppet Extension Version:**master (6886aaf)
+ **Puppet Agent Version:** 5.3.2 (`puppet --version`)
+ **PDK Version (if applicable):** Not installed.

### What You Are Seeing?
When specifying a file resource and checking the node graph, backslashes are not displayed correctly, though forward slashes are.

![image](https://user-images.githubusercontent.com/14190564/32082630-8b081ee6-ba82-11e7-8c66-aa05b2740400.png)

![image](https://user-images.githubusercontent.com/14190564/32082634-9526b5ea-ba82-11e7-8c56-8586320cdb03.png)

![image](https://user-images.githubusercontent.com/14190564/32082640-9e6ec430-ba82-11e7-91eb-de9b8aecc1d6.png)

### What is Expected?
Paths to appear correctly in the node graph.

### How Did You Get This To Happen? (Steps to Reproduce)
Clone, build, use the code from the screen shots.

### Output Log
None generated.

glennsarti:
So after some poking around.  In the DOT language,  backslashes in the label of a node a allowed and should not be seen as escaping characters.  However renderers may, and in this case I believe that's what is happening.  So this means the backslash munging needs to go into the client side, not language server as the server is outputting a valid DOT file.

glennsarti:
@michaeltlombardi  Happy to close this issue?

michaeltlombardi:
Yes!

PR #185, opened 2017-10-30
(GH-180) Display backslashes in node graph

glennsarti:
Previously vis.jz sees backslashes as escape characters, however they are not in
the DOT language.  This commit escapes any backslash (double backslash) coming
from a valid DOT file, in preparation to be rendered by viz.js.


glennsarti:
Confirmed that backslashes work in the PE Console, so it's definitely viz.js not rendering correctly

jpogran:
@michaeltlombardi has verified this works!

PR #186, opened 2017-10-31
(maint) Fix rubocop violations

glennsarti:
Fixes rubocop violations


glennsarti:
Verified manually that the language server still works as expected on Windows.

Issue #187, opened 2017-11-02
Enhancement: Add support for an --stdio mode on the language server


ace13:
Currently, the language server only supports running beneath a TCP socket, but there are language server protocol clients which expect to be able to run a server and use direct stdio communication instead of having to handle sockets.

glennsarti:
Merged and closing.

PR #188, opened 2017-11-06
(GH-#187) Add a stdio mode to the language server

ace13:
I'm unsure on how to add a test for this feature, but if someone has a good suggestion for that then I'll happily write one.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -111,6 +111,7 @@ Usage: puppet-languageserver.rb [options]
     -d, --no-preload                 Do not preload Puppet information when the language server starts.  Default is to preload
         --debug=DEBUG                Output debug information.  Either specify a filename or 'STDOUT'.  Default is no debug output
     -s, --slow-start                 Delay starting the TCP Server until Puppet initialisation has completed.  Default is to start fast
+        --stdio                      Runs the server in stdio mode, without a TCP listener
     -h, --help                       Prints this help
     -v, --version                    Prints the Langauge Server version
 ```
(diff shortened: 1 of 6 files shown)


glennsarti:
Thanks for the PR @ace13 .  I'll review it soon.

glennsarti:
@ace13  Are you ok with me adding on some commits to this PR?

glennsarti:
Couple of things I see
* Client detection is a bit off so the timeout doesn't work yet
* I'll have a go at some basic tests for both stdio and tcp based servers
* I want to add a failure if you specify --stdio and --debug=STDOUT as they're not compatible

ace13:
I didn't touch any of that client detection, since there's technically a client through the entirety of the servers lifetime. Not sure if that's the right way to go about it though, or if actually tracking a client is more appropriate.

I also did add a small mode for specifying `--debug=STDERR` during my testing, but didn't want to commit it as part of this feature addition as it's completely irrelevant to the feature itself. But yeah, should probably have an error for debug and stdio mode at the same time. 

And I don't particularly mind if you want to add some commits as well.

glennsarti:
Yeah...just trying to think, if the language server is started as a child process, and the parent dies, then the child is orphaned.  Will the language server be able to detect that ... I'll see if I can set that up as testing scenario.

glennsarti:
Yup...the language server remains if the parent dies.  At least on Windows :-(

glennsarti:
On second thoughts.  I think I'm being too defensive.  Will just write some tests and then merge it.

glennsarti:
Hrmm...so the TCP server code seems to be broken.  Good news is, my tests were failing because of it!  I'll keep working on it.  Thanks @ace13

glennsarti:
@ace13 You ok with my additional commits?

ace13:
No issues with them, looks fine to me.

PR #189, opened 2017-11-10
WIP: MkDocs Pilot

michaeltlombardi:
Docs pilot for building out docs without copy-pasta / duplicating assets. Includes structure but is very rough; CI to follow. Need to decide if we want to build in bash or PowerShell or both?

Also, there's a [live preview](https://michaeltlombardi.github.io/puppet-vscode/), but hasn't been themed up properly.


michaeltlombardi:
`docs.sh` is currently **very unsafe** to run, just using it to nail down what needs doing. It doesn't protect you from not having the software or from overwriting the config multiple times.

michaeltlombardi:
The commit message is also placeholder nonsense.

.gitignore
```diff
@@ -0,0 +1,3 @@
+docs/about
+docs/img/logo.png
```
glennsarti: Oddly specific ignore...

michaeltlombardi: That's because, in this incarnation, we are converting the SVG to PNG for use as the docs logo (it doesn't currently accept SVG, will look to fix or find alternative when we get around to  theming). 

Reference:

https://github.com/jpogran/puppet-vscode/pull/189/files#diff-270eb1b0d54227affb1d5828212609ddR16

michaeltlombardi: We probably _don't_ want to ignore **all** future image assets though.

jpogran:
Closing with intent to implement in new repo organization

Issue #166, opened 2017-10-12
Using f12 / ctrl f12


jgrammen-agilitypr:
Neither f12 nor ctrl f12 appear to work in visual studio code with the 0.7.1 version of the extension installed.
I cannot select a class (eg: users::my_name ) and press a key to go to the implementation/definition of that class. Should this work? maybe i am using the wrong shortcut?

glennsarti:
Hi @jgrammen-agilitypr .  The extension has never supported the peek definition feature.  What I will do is add that as a feature request.

jgrammen-agilitypr:
Hello, Thank you for the reply.
Ok, yes making it a feature request sounds about right.
It certainly would be very useful.
I currently use Geppetto and being able to jump to declaration / definition is something I use frequently.

glennsarti:
Ahh ok.  This may require some fiddling within Puppet to get that information but in theory it sounds like it's possible.

jgrammen-agilitypr:
In case my feature request was not completely clear, hear are a few images:
![gotodeclaration](https://user-images.githubusercontent.com/22965146/31560648-20507e5e-b023-11e7-90fc-1afb484d44c7.png)
![gotodeclarationf2](https://user-images.githubusercontent.com/22965146/31560649-2060b922-b023-11e7-944a-ff82aad4824f.png)
When you click on the button it opens the file that defines Class mm::motd

glennsarti:
Yup that's what I'm thinking of:

https://code.visualstudio.com/docs/editor/editingevolved#_go-to-definition

https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#goto-definition-request

mdwheele:
@glennsarti, do you think this is something achievable without a ton of experience / intimate knowledge of VS Code extensions? I would be willing to make an attempt if you could point me in the right direction to get started.

glennsarti:
Responding to requests from VScode and crafting the write JSON messages (as per language server protocol is easy enough).

The thing I don't know is;

Does puppet keep a track of which puppet classes were loaded from which file?
Does puppet keep track of which puppet functions were loaded from which file?
If that's stored somewhere then wiring that up isn't too hard.
If that's _not_ stored that will require monkey patching the loaders in Puppet, and THAT is not a simple thing to do.

jgrammen-agilitypr:
any chance on this getting merged in? or is it still a work in progress?

glennsarti:
[...]

PR #193, opened 2017-11-15
(GH-166) Add find/peek definition capability to language server

glennsarti:
Previously the Language Server did not respond to find definition requests from
the client.  This commit
* Adds a request handler for the textDocument/definition request
* A definition provider class which will extract the definition
* Monkey patch Type and Function loading so the language server can determine
  where types and functions came from
* Added definition provider as a language server capability

This commit adds parsing of puppet manifests for classes and defined types for
manifests that exist in the module path.
* The classes are loading async if preload is configured
* Will respond to definition requests if a type with the same name could not be
  found

---

This commit adds class loading the client UI so instead of steps 1 to 3, there
are now 4 steps (Functions, Classes, Types and Facts).

---

reviously definition detection would fail on resource bodies as the text comes
in as a LiteralString e.g.

class { 'testclass':    <--- testclass would be the LiteralString inside
} 							 a ResourceBody

This commit updates the definition provider to detect these instances and return
a defintion location.  This commit also adds tests for this scenario.

---

Note this does not support finding definitions for things in the current workspace, only in the modulepath or default puppet locations.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -33,6 +33,7 @@ Create or open any Puppet manifest with the extension `.pp` or `.epp` and the ex
 - Code snippets
 - Linting
 - IntelliSense for resources, parameters and more
+- Go to Definition of functions, types and classes
 - Validation of `metadata.json` files
 - Import from `puppet resource` directly into manifests
 - Node graph preview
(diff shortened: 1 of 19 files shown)


glennsarti:
Rebased after all the test file locations changed.

Issue #204, opened 2017-12-20
Puppet Debugger Issue


dotanspector:
### What Versions are you running?

OS Version:  Mac OSX Sierra
VSCode Version: 1.18.1
Puppet Extension Version: 0.8.0
Puppet Agent Version: 4.10.1
PDK Version (if applicable):1.2.1

### What You Are Seeing?
Im Trying to run the debugger and encounter this message:
   "Err: Evaluation Error: Error while evaluating an Object, undefined method `file' for 
     Puppet::Pops::Model::HostClassDefinition:Puppet::Pops::Model::HostClassDefinition on node 
     <my_node>
     puppet exited with 1"
### What is Expected?
successful run
### How Did You Get This To Happen? (Steps to Reproduce)
debugger configuration:     "configurations": [
        {
            "type": "Puppet",
            "request": "launch",
            "name": "Puppet Apply current file",
            "manifest": "${file}",
            "args": [],
            "noop": true,
            "cwd": "${file}"
        }
    ]
}

glennsarti:
@dotanspector Hi.  Thanks for the bug report.  Are you able to copy the manifest which caused the issue?

dotanspector:
Hi @glennsarti ,
Thanks in advance for your support
**This is the manifest (dummy manifest)**

```
# dotantest
#
# A description of what this class does
#
# @summary A short summary of the purpose of this class
#
# @example
#   include dotantest
class dotantest {
  user { 'title':
    ensure => 'present'
  }

}
```

**Debugger output**

```
**************************************************
* The Puppet debugger is an experimental feature *
* Debug Server v0.0.1                            *
**************************************************
puppet apply /Users/dotan.spector/<path_to_manifest>/dotantest/manifests/init.pp --detailed-exitcodes --logdest debugserver --noop --trace
Warning: Facter: effective gid 1371015622 does not have a group entry.
Err: Evaluation Error: Error while evaluating an Object, undefined method `file' for Puppet::Pops::Model::HostClassDefinition:Puppet::Pops::Model::HostClassDefinition on node dspector-mc.local
/Users/dotan.spector/.vscode/extensions/jpogran.puppet-vscode-0.8.0/vendor/languageserver/lib/puppet-debugserver/debug_hook_handlers.rb:31:in `hook_before_pops_evaluate'
[...]

PR #205, opened 2017-12-23
(GH-204) Fix debug server for Puppet 4.x

glennsarti:
Previously it was hoped to have the tests for the debug and language servers in
the same directory however as the two applications are separate and have very
distinct setups, they needed to be split.  This commit splits out the language
server tests into their own directory and modifies the Rakefile to be able to
invoke either set of tests.

Note the --default-path parameter will now be needed when running rspec directly
for example;

rspec --default-path ./spec/debugserver spec/debugserver/unit

This commit also modifies the appveyor.yaml file to run both the language server
and debug server tests in all modern puppet gem versions.

Previously the Debug Server only support Puppet 5 agents.  However a large number
of users still run Puppet 4.x.  This commit:

* Refactors function calls which can differ between Puppet 4 and 5 into helper
  methods. Namely positioning information and class name shortcuts.

* Identifies other potential issues which use Puppet 5 only concepts


dotanspector:
Hi @glennsarti 
Thank you very much.
Waiting for Merging

PR #213, opened 2018-01-29
Add debug server integration tests

glennsarti:
Previously the evaluate hooks would be firing when the debug session was paused.
This means when doing string evaluations the POPS evaluation depth would be
modified and potentially other hooks (function, exception) would be firing in
error.  This commit guards the evaluation hooks to exit early if the session is
paused.

This commit also fixes a bug in the after evaluation hook which was decrementing
the counter back to zero, which then broke the Step Out functionality.  This
commit changes the functionality to only decrement by one per invocation.

---

Previously the sequence number was doubling instead of incrementing by one. This
was due to a previous typo.  This commit changes the increment to one.

---

Previously, generating the stack frame list was raising errors due to Puppet 4
not exposing the same methods.  This was mostly fixed in 6740faf however
the stack frames were not testable.  This commit changes the location detection
to use the same helper methods created in 6740faf and also adds the
pos_on_line helper method.

---

Previously due to puppetlabs/puppet@0f96dd9
the stack trace from puppet was being output in the reverse order.  This commit
reverts the order for Puppet 4.7.x and below, to mirror that of version 4.8.0
and above.

---

Previously the test coverage on the debug server was minimal.  This commit adds
integration tests for the debug server.  In particular it starts a debug server
as a separate process and emulates debug client TCP messages.  The kitchen sink
test file, tests all of the features of the debug server.

### DIFF
--- a/server/lib/debugserver/debug_protocol.rb
+++ b/server/lib/debugserver/debug_protocol.rb
@@ -334,6 +334,15 @@ module PuppetDebugServer
       end
     end
 
+    # /** Response to 'configurationDone' request. This is just an acknowledgement, so no body field is required. */
+    # export interface ConfigurationDoneResponse extends Response {
+    # }
+    module ConfigurationDoneResponse
+      def self.create_from_request(options, request = nil)
+        Response.create_from_request(options, request)
+      end
+    end
+
     # /** Launch request; value of command field is 'launch'. */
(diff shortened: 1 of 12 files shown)


glennsarti:
Found a bug in the debug server for Puppet 4.x.  5.x is fine.

glennsarti:
Bounced it to force a build

PR #242, opened 2018-03-26
(WIP)(maint) Fix build errors

jpogran:
This commit updates tsconfig.json to point to the src folder and not the root directory. It also points the module import for viz.js to the file instead of folder to avoid build errors


glennsarti:
Is this still WIP?

jpogran:
Closing as this is not needed after all.

Issue #252, opened 2018-04-04
Org Move: Update README


jpogran:
All links in the readme that point to `jpogran/puppet-vscode` need to be changed to `lingua-pupuli`.

Some links have `puppet-vscode/master/client` in them, which will need `client` removed after PR #251 is merged.

Issue #250, opened 2018-04-04
Org Move: Move client folder to root


jpogran:
As part of the move of this repo to the lingua-pupuli org, the language server and the client extension have been split into separate repos.

This ticket tracks the work to move the contents of the client folder to the root of the project folder heirachy

PR #255, opened 2018-04-04
(GH-252) Update README for Org Move

jpogran:
This commit updated all references from `jpogran/puppet-vscode` to `lingua-pupuli` as well as removing any references to the `client` folder which was removed in GH-250

PR #251 must be merged first


glennsarti:
Rebase required.

glennsarti:
Merging as CI is not required for README only change.

PR #272, opened 2018-05-16
Refactor Configuration

jpogran:
This PR refactors the confgiruation and invocation of the LangugeServerClient and the LanguageServer. In doing so, it enables starting the LanguageServer using `stdio` instead of `tcp`.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1,28 +1,164 @@
 'use strict';
 
 import * as vscode from 'vscode';
+import * as path from 'path';
 
 import { IConnectionConfiguration, ConnectionType } from './interfaces';
 
 export class ConnectionConfiguration implements IConnectionConfiguration {
-  public type: ConnectionType = ConnectionType.Unknown;
   public host: string;
   public port: number;
   public timeout: number;
   public enableFileCache: boolean;
   public debugFilePath: string;
-  public puppetAgentDir: string;
   public langID: string = 'puppet'; // don't change this
+  config:vscode.WorkspaceConfiguration;
+  context:vscode.ExtensionContext;
 
   constructor(context: vscode.ExtensionContext) {
-    let config = vscode.workspace.getConfiguration('puppet');
+    this.context = context;
+    this.config = vscode.workspace.getConfiguration('puppet');
 
-    this.host = config['languageserver']['address'];
-    this.port = config['languageserver']['port'];
-    this.timeout = config['languageserver']['timeout'];
-    this.enableFileCache = config['languageserver']['filecache']['enable'];
-    this.debugFilePath = config['languageserver']['debugFilePath'];
+    this.host = this.config['languageserver']['address'];
+    this.port = this.config['languageserver']['port'];
+    this.timeout = this.config['languageserver']['timeout'];
+    this.enableFileCache = this.config['languageserver']['filecache']['enable'];
+    this.debugFilePath = this.config['languageserver']['debugFilePath'];
+  }
 
-    this.puppetAgentDir = config['puppetAgentDir'];
+  get puppetAgentDir():string{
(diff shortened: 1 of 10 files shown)


src/debugAdapter.ts
```diff
@@ -77,6 +78,7 @@ class DebugConfiguration implements IConnectionConfiguration {
     this.sslCertDir ='';
     this.pdkDir ='';
     this.languageServerCommandLine =[''];
+    this.protocol = ProtocolType.STDIO;
```
glennsarti: nit: Debug server only supports TCP but the code never looks at this property

src/configuration.ts
```diff
@@ -1,28 +1,219 @@
...
+    if (this.host === '127.0.0.1' || this.host === 'localhost' || this.host === '') {
+      return ConnectionType.Local;
+    } else {
+      return ConnectionType.Remote;
+    }
+  }
 
-    this.puppetAgentDir = config['puppetAgentDir'];
+  get protocol(): ProtocolType {
+    switch (this.config['languageserver']['protocol']) {
```
michaeltlombardi: Here we specify the config as `puppet.languageserver.protocol` but in the [package.json](https://github.com/lingua-pupuli/puppet-vscode/pull/272/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R275) it is defined as `puppet.languageclient.protocol`. Both of these setting appear to be new and should map correctly - this currently causes connecting to a remote language server over TCP to fail.

michaeltlombardi:
Debug flat out blew up at me with a `Debug adapter process has terminated unexpectedly` error. 😞

src/configuration.ts
jpogran: Fixed. Good catch, the default case statement catches the error and starts the STDIO server which is the only thing I cared about at the time. I think these setting names need a review, if a person familiar with them can't keep them straight :)

jpogran:
@michaeltlombardi debugadapter fixed. Still does not break at a breakpoint, but that is a known issue

glennsarti:
Hrmmm weird.  I can't get the debug adapter to work after commit 4aaf3c516704a06e2fb0ef6f038f4619dc4509ff was merged.  More investigation required.

That shouldn't stop this PR being merged, but it needs to be fixed before the next release.

jpogran:
@glennsarti That is a good catch! However it's indicative that these shouldn't share configuration classes/interfaces, there are too many differences and/or incompatibilities. We should split these up, and anything that uses them should know how to pull in the correct pieces. Will work on that after fixing the import

src/debugAdapter.ts
```diff
@@ -5,7 +5,7 @@ import path = require('path');
 import net = require('net');
 import { DebugProtocol } from 'vscode-debugprotocol';
 import cp = require('child_process');
-import { NullLogger } from './logging/null';
+import { FileLogger } from './logging/file';
```
glennsarti: The logging needs to remain as Null Logger for the time being.  There's no error checking of the log file exists so it blows up the DebugAdapter.  If we really want the file logger at least comment it out

`// import { FileLogger } from './logging/file';`

src/debugAdapter.ts
```diff
@@ -187,11 +198,11 @@ function startDebugging(config:DebugConfiguration, debugLogger:ILogger) {
 }
 
 // TODO Do we need a logger? should it be optional?
-// var logPath = path.resolve(__dirname, "../logs");
-// var logFile = path.resolve(logPath,"DebugAdapter.log");
-// let debugLogger = new FileLogger(logFile);
+var logPath = path.resolve(__dirname, "../logs");
```
glennsarti: See comments above about keeping the logger as Null for the time being.

Issue #274, opened 2018-05-22
Remove random port detection


jpogran:
Currently the extension determines a random unused port in order to communicate with the language server over tcp. In PR #279 stdio will become the default protocol, meaning using tcp will be mostly used by those doing advanced scenarios where they will be picking the host name and port and won't want something attempting to pick a port for them. 

This issue suggests we remove the random port detection and rely on the configured port to connect if using tcp. This would avoid some of the problems in #240, while also addressing some of it.

PR #294, opened 2018-06-21
(GH-274) Remove random port detection

jpogran:


src/connection.ts
```diff
@@ -127,8 +127,14 @@ export class ConnectionManager implements IConnectionManager {
           this.logger.debug('OUTPUT: ' + data.toString());
 
           // If the language client isn't already running and it's sent the trigger text, start up a client
-          if (this.languageServerClient === undefined && data.toString().match('LANGUAGE SERVER RUNNING') !== null) {
-            this.languageServerClient = this.createLanguageClient();
+          if (this.languageServerClient === undefined && /LANGUAGE SERVER RUNNING/.test(data.toString())) {
+            if(this.connectionConfiguration.port){            
+              this.languageServerClient = this.createLanguageClient();
+            }else{
+              var p = data.toString().match(/LANGUAGE SERVER RUNNING \d+\.\d+\.\d+\.\d+:(\d+)/);
```
glennsarti: The hostname can be text e.g. `localhost`. so the IP based regex will fail.  Perhaps

`/LANGUAGE SERVER RUNNING .+:(\d+)$/` ?

jpogran: Good point, and I wasn't even using that match. Though, I think its `.*` isn't it?

jpogran:
ping @glennsarti

Issue #240, opened 2018-03-21
TCP Lang client should have reasonable retry defaults


glennsarti:
In #220 it was found that the TCP lang client connector is not using any sane TCP connect retry options which can cause really bad sideeffects.

This issue should resolve;
* A delay between retry
* A max attempt count
* An actual UI Warning then Error message when it cannot bind a port

glennsarti:
Merged

PR #297, opened 2018-06-25
(GH-240) TCP Retry defaults

jpogran:
This implements tcp retry defaults for the languageclient

- [x] A delay between retry
- [x] A max attempt count
- [x] An actual UI Warning then Error message when it cannot bind a port

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -228,19 +228,65 @@ export class ConnectionManager implements IConnectionManager {
   ): ServerOptions {
     let serverOptions: ServerOptions = function() {
       return new Promise((resolve, reject) => {
+        const retries = 5;
+        var attempt = 0;
         var client = new net.Socket();
+
+        const rconnect = () => { client.connect(port, address) };
+
         client.connect(port, address, function() {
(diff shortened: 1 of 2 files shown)


jpogran:
Added UI warning message.

image

jpogran:
Error colllapsed:
image

Error expanded:
image

src/connection.ts
```diff
@@ -228,19 +228,65 @@ export class ConnectionManager implements IConnectionManager {
...
+              }
+            );
+
+            return null;
+          }else{
+            attempt = attempt + 1;
+            var message = `Timed out connecting to language server. Is the server running at ${address}:${port} ? Will wait timeout value before trying again`;
+            switch(err.code){
+              case 'ETIMEDOUT':
+                message = `Timed out connecting to language server. Is the server running at ${address}:${port} ? Will wait timeout value before trying again`;
```
glennsarti: Should `Will wait timeout value before trying again` be `Will wait timeout 5 seconds before trying again`

jpogran: The timeout for ETIMEOUT is ~30 seconds. I didn’t see a method to change that

Issue #301, opened 2018-06-27
Extension fails silently when no puppet agent is found


jpogran:
If the extension can't find a Puppet Agent installation or was configured with an incorrect path, the extension silently fails to load the language server and there isn't much indication you're in a broken state. I suggest it should attempt to find the puppet binary almost immediately and fail with an error message in the UI.

PR #302, opened 2018-06-27
(GH-301) Fail fast if no agent installed

jpogran:
This PR adds logic to check if there is a valid Puppet Agent binary
file path configured at extension load. It will then throw an
informative error if there isn't a binary found so the user knows what
won't work without an agent.

Previously if the extension couldn't find a Puppet Agent installation or
was configured with an incorrect path, the extension silently failed to
load the language server and there wasn't much indication you're in a
broken state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ get notified automatically about any future extension updates!
 
 ## Experience a Problem?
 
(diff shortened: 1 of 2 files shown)


jpogran:
image

Issue #307, opened 2018-07-11
Path auto-detection does not work on *nix and mac


jpogran:
With the new configuration class, resolving paths on *nix and mac installs does not work.

The install path for Puppet in Windows is `$env:ProgramFiles\Puppet Labs\Puppet`, but in the Mac and *nix packages it's `/opt/puppetlabs` (unless you're arch linux where it's `/usr/bin/puppet` but let's ignore that for now). This is important because the extension previously assumed it was `/opt/puppetlabs/puppet` on *nix and Mac, which would then be resolved to `/opt/puppetlabs/puppet/puppet` when the extension later on resolved the directory for the Puppet binaries.

glennsarti:
Closed by merge

PR #308, opened 2018-07-11
(GH-307) Fix path resolution on mac and *nix

jpogran:


glennsarti:
So I tested this on my Ubuntu desktop and it worked.  Although it wasn't broken before.  Did we expect all *nix users to be broken?

jpogran:
I updated the issue with a little more information

src/configuration.ts
```diff
@@ -26,7 +26,7 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
     this.debugFilePath = this.config['languageserver']['debugFilePath'];
   }
 
-  get puppetAgentDir(): string {
+  get puppetBaseDir(): string {
```
jpogran: I didn't have to change the name of the accessor here, but it helped keep things straight in my head.

src/configuration.ts
```diff
@@ -37,9 +37,12 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
         if (process.env['PROCESSOR_ARCHITEW6432'] === 'AMD64') {
           programFiles = process.env['ProgramW6432'] || 'C:\\Program Files';
         }
+        // On Windows we have a subfolder called 'Puppet' that has
+        // every product underneath
         return path.join(programFiles, 'Puppet Labs', 'Puppet');
       default:
-        return '/opt/puppetlabs/puppet';
+        // On *nix we don't have a sub folder called 'Puppet'
+        return '/opt/puppetlabs';
```
jpogran: This was `/opt/puppetlabs/puppet` before, which would make `puppetDir` `/opt/puppetlabs/puppet/puppet` which doesn't exist

src/configuration.ts
```diff
@@ -61,34 +64,45 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
...
+    return path.join(this.puppetBaseDir, 'mcollective');
   }
 
   get rubydir(): string {
-    return path.join(this.puppetAgentDir, 'sys', 'ruby');
+    switch(process.platform){
+      case 'win32':
+        return path.join(this.puppetBaseDir, 'sys', 'ruby');
+      default:
+        return path.join(this.puppetBaseDir, 'lib', 'ruby');
```
jpogran: The vendored ruby paths are different on Mac and *nix, so use the correct path here

src/configuration.ts
```diff
@@ -61,34 +64,45 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
...
-      path.join(this.facterDir, 'lib') +
-      this.pathEnvSeparator() +
-      path.join(this.hieraDir, 'lib') +
-      this.pathEnvSeparator() +
-      path.join(this.mcoDir, 'lib');
+    var p = new Array(
+      path.join(this.puppetDir, 'lib'),
+      path.join(this.facterDir, 'lib'),
+      // path.join(this.hieraDir, 'lib'),
+    ).join(this.pathEnvSeparator());
```
jpogran: Reduced the amount of lines needed for this by using an Array.join

src/configuration.ts
```diff
@@ -98,31 +112,16 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
...
-      this.pathEnvSeparator()
-    );
+    return new Array(
+      path.join(this.puppetDir, 'bin'),
+      path.join(this.facterDir, 'bin'),
+      // path.join(this.hieraDir, 'bin'),
+      path.join(this.puppetBaseDir, 'bin'),
+      path.join(this.rubydir, 'bin'),
+      path.join(this.puppetBaseDir, 'sys', 'tools', 'bin')
+    ).join(this.pathEnvSeparator());
```
jpogran: Reduced the amount of lines needed for this by using an Array.join

src/rubyHelper.ts
```diff
@@ -52,7 +52,7 @@ export class RubyHelper {
       result.options.env.RUBYLIB = '';
     }
     result.options.env.RUBY_DIR = connectionConfiguration.rubydir;
-    result.options.env.PATH =connectionConfiguration.environmentPath + result.options.env.PATH;
+    result.options.env.PATH =connectionConfiguration.environmentPath + this.pathEnvSeparator() + result.options.env.PATH;
```
jpogran: The missing path separator caused problems

glennsarti:
Seems Arch is supported by VSCode...https://aur.archlinux.org/packages/visual-studio-code-bin/?comments=all

src/rubyHelper.ts
glennsarti: nit: should there be a space between `=connectionC`

glennsarti:
It's odd.  I can't replicate the bad paths with my Ubuntu Desktop.  Another user I had try with Mac is not experiencing the bad pathing.  And I can't see anything in the Puppet repos/docs to suggest a path change.  This PR looks fine and I'm almost about to press the merge button, but I'm having trouble repro-ing.

`no issues from what I can see. VSC 1.25.0, PE 4.10.9. im on OSX 10.12.6, so not high sierra`

PR #311, opened 2018-07-12
(GH-310) Fix gulp bump

### LINKED ISSUE #310: Appveyor build does not version vsix correctly
Instead of bumping the version in package.json so that the resulting vsix is named correctly, gulp bump outputs '3.9.1'. This results in a vsix that still uses the production released version, which makes using a 'nightly' build harder than it should

```
[GH-307-fix-path-resolution-on-nix-and-mac +0 ~3 -0 !] ~\src\lingua-pupuli\puppet-vscode
07-11 15:07:56 30> node .\node_modules\gulp\bin\gulp.js bump --version '99.99.0-appv.127'
3.9.1
```

jpogran:
This PR fixes bumping the version of the extension in package.json by changing the parameter name used to specify version. It appears version is already used by the yargs package (https://github.com/yargs/yargs/blob/master/docs/api.md#versionoption-description-version) which returns the version of the yargs package instead of what you supplied.

This PR successfully makes a versioned vsix in appveyor: https://ci.appveyor.com/api/buildjobs/yeg6k3q119arvqwn/artifacts/puppet-vscode-99.99.0-appv.134.vsix

Fixes #310

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -40,6 +40,17 @@
         "${workspaceFolder}/out/test/**/*.js"
       ],
       "preLaunchTask": "npm: watch"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Gulp task",
+      "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+      "args": [
+        // Add arguments here to debug gulp tasks
+        // "bump", "--type", "minor"
+        // "bump", "--specific", "99.99.0-appv.127"
+      ]
     }
   ],
   "compounds": [
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,7 @@ install:
        & npm install --silent
 
        Write-Host "Set the package.json version..."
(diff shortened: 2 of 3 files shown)


.vscode/launch.json
```diff
@@ -40,6 +40,16 @@
...
       "preLaunchTask": "npm: watch"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Gulp task",
+      "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
+      "args": [
+        // "bump", "--type", "minor"
+        "bump", "--specific", "99.99.0-appv.127"
```
glennsarti: Should this be here?

jpogran: I thought to include this, as it was a way to debug gulp tasks using vscode. If you don't want the example in there, I can take it out

glennsarti: Ahhh Ok.  I was a confused because of the very specific version number instead of asking the user.

jpogran: Commented out the specific lines and added a short comment explaining how to use.

Issue #319, opened 2018-07-17
IF/ELSE Snippet has incorrect increment


jpogran:
The snippet for if/else incorrectly shows the wrong comment:

```
if condition {
  # when true
}
else {
  # when true
}
```

It should show:

```
if condition {
  # when true
}
else {
  # when false
}
```

This is because line https://github.com/lingua-pupuli/puppet-vscode/blob/master/snippets/keywords.snippets.json#L9 should have a 3 not a 2.

jpogran:
Done

PR #320, opened 2018-07-17
(GH-319) Fix if/else snippet

jpogran:
This commit fixes the if/else snippet by entering in the correct
increment in the tab stop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## Unreleased
 
+- ([GH-320](https://github.com/lingua-pupuli/puppet-vscode/issues/320)) Fix if/else snippet
```
michaeltlombardi: Might be useful to include the heading it will eventually go under after we update the subheadings for prior releases. Shouldn't hold up this PR.

jpogran: Good point. Done

PR #325, opened 2018-07-19
(GH-317) Fix extension test harness

### LINKED ISSUE #317: Fix extension test harness
The current test harness is pointed at the wrong directory `out\test`, when it should be `out\src\test`

jpogran:
This PR fixes the test harness to actually run both interactively and from the command line. This resolves #317 

To test (run `git clean -xfd` before each step):

- Run `npm test` without VSCode running
- Run `Extension Host` launch task
- Run `Extension Tests` launch task

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "onCommand:extension.pdkTestUnit",
     "onCommand:extension.pdkValidate"
   ],
-  "main": "./out/src/extension",
+  "main": "./out/extension",
   "contributes": {
     "languages": [
       {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -258,7 +258,7 @@ export class ConnectionManager implements IConnectionManager {
           }else{
             attempt = attempt + 1;
             var message = `Timed out connecting to language server. Is the server running at ${address}:${port} ? Will wait timeout value before trying again`;
-            switch(err.code){
+            switch(err['code']){
               case 'ETIMEDOUT':
                 message = `Timed out connecting to language server. Is the server running at ${address}:${port} ? Will wait timeout value before trying again`;
                 break;
--- a/src/debugAdapter.ts
+++ b/src/debugAdapter.ts
@@ -39,7 +39,7 @@ function sendErrorMessage(message: string) {
 
 class DebugConfiguration implements IConnectionConfiguration {
   protocol: ProtocolType;
-  puppetAgentDir: string;
+  puppetBaseDir: string;
   languageServerPath: string;
   rubydir: string;
   rubylib: string;
@@ -62,7 +62,7 @@ class DebugConfiguration implements IConnectionConfiguration {
     this.port = 8082;
     this.timeout = 10;
     
-    this.puppetAgentDir = '';
+    this.puppetBaseDir = '';
     this.languageServerPath = '';
     this.rubydir ='';
     this.rubylib ='';
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "module": "commonjs",
     "target": "es6",
-    "outDir": "out/src",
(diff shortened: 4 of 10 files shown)


src/connection.ts
```diff
@@ -258,7 +258,7 @@ export class ConnectionManager implements IConnectionManager {
           }else{
             attempt = attempt + 1;
             var message = `Timed out connecting to language server. Is the server running at ${address}:${port} ? Will wait timeout value before trying again`;
-            switch(err.code){
+            switch(err['code']){
```
jpogran: This is an object and not a class, there isn't technically a property called `code`. This is essentially the same call, but is accepted by tslint.

src/debugAdapter.ts
```diff
@@ -39,7 +39,7 @@ function sendErrorMessage(message: string) {
 
 class DebugConfiguration implements IConnectionConfiguration {
   protocol: ProtocolType;
-  puppetAgentDir: string;
+  puppetBaseDir: string;
```
jpogran: Missed rename from #307

tsconfig.json
```diff
@@ -1,23 +1,23 @@
...
+      "module": "commonjs",
+      "target": "es6",
+      "outDir": "out",
+      "lib": [
+          "es6"
+      ],
+      "sourceMap": true,
+      "rootDir": "src",
+      /* Strict Type-Checking Option */
+      "strict": false,   /* enable all strict type-checking options */
```
jpogran: This is necessary to make the test runs green. There are far too many `strict` errors to fix in this PR, so we should accept the tech debt and turn off strict since we can't address the suggestions now. I will make a ticket to come back and address them

jpogran:
Still need to clean up commits and discuss whether we disable tslint `strict` or not

glennsarti:
@jpogran I might do up PR to add a step to Appveyor or Travis to run tslint with Strict but expect failures.  Seem as though we can't do the equivalent of `.rubocop_todo.yaml` in tslint.

package.json
```diff
@@ -392,7 +392,7 @@
   "scripts": {
     "vscode:prepublish": "node node_modules/gulp/bin/gulp.js build",
     "compile": "tsc -p ./",
-    "watch": "node node_modules/gulp/bin/gulp.js initial && tsc -watch -p ./",
```
glennsarti: The gulp task was used to auto-vendor editor-services first before running tsc.  Not sure why it was removed.

jpogran:
@glennsarti so I've done some more digging, and there is two challenges here.

The goal of this PR is to enable running the tests inside VSCode using the `Extension Tests` launcher, or using `npm test`. After fixing the harness directory part, everything built, but tests would still refuse to run because of typescript errors. I have always thought there were tslint errors that I hadn't figured out the correct way to ignore yet. After actually looking at things I found that there are two types of errors we have in the project: tsc errors, and tslint errors.

The tslint errors don't prevent the tests from running and are mostly stylistic or anti-patterns stuff at this point in the project.

The tsc errors are compile time errors, and therefore won't be ignored by tslint (which explains my failed attempts using `ts-lint:disable`. To ignore these errors we must add `@ts-ignore` to every line, but this ignore all errors for that line not just the error we want. It also comes with a heavy warning to not use from the compiler team as it is a bad thing. 

In short, there isn't a `expect errors` as these are compile time errors and would prevent compilation if this were a static typed language like C#. The way to truly `expect errors` is to disable strict, which is what part of this PR does.

jpogran:
So it seems like the choice is, one global on/off switch that can be easily found, or multiple disable lines all over the project. I vote for the one global, and a ticket to address

michaeltlombardi:
I'm in favor of the global switch and tickets to address individual issues.

michaeltlombardi:
shipit

Issue #315, opened 2018-07-17
Abstract Puppet and Ruby path detection to allow PDK use


jpogran:
There was alot of work done in https://github.com/lingua-pupuli/puppet-vscode/pull/272, among other commits to improve the detection and configuration of settings needed for the LanguageServer to run and allow us to change things like TCP or STDIO without changing half the code base. However it stopped short of enabled us to choose PDK or Puppet-Agent as the Puppet and Ruby source easily in a supportable way.

This issue will refactor the path detection and resolution for the Puppet and Ruby paths to allow using the PDK as a source instead of the Puppet Agent.

It will:

- If a user specifies a custom path, it will use that path and stop resolution
- If a user does not specify a custom path, it will attempt to use PDK first
- If PDK is not found, it will attempt to use Puppet Agent
- If no valid path is found, it will throw an error and not attempt to load the language server

Note, that in order to do this we have to account for the different installation paths in Windows, Linux, and Mac *and* the different paths between PDK and Puppet-Agent.

glennsarti:
Fixed by https://github.com/lingua-pupuli/puppet-vscode/pull/330

PR #330, opened 2018-07-26
(GH-315) PDKIFY ALL THE THINGS

jpogran:
This PR adds PDK as a source location for the language server.

- A new configuration point is added: `puppet.installType` which reflects either PDK or Puppet-Agent, default set to `pdk`.
- A new configuration point is added: `puppet.InstallDirectory` which can refer to either a Puppet-Agent directory or a PDK directory.
- The `puppet.puppetAgentDir` configuration point has been removed

If no configuration is set by the user, the extension will look for a PDK install in the default location.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,7 @@ language: cpp
 
 os:
   - linux
-# Do we actually need to run this on OSX?
-# OSX is REALLY slow to test
-#  - osx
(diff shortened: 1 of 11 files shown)


glennsarti:
Hrmmm.  Still can't get the debug adapter working yet...

PR #333, opened 2018-07-31
(maint) Update Node and Typescript versions

jpogran:


jpogran:
@glennsarti @michaeltlombardi would you think updating node and typescript versions requires a ticket number?

.travis.yml
```diff
@@ -18,7 +18,7 @@ env:
     # Extension testing
 
     # Package it up to a VSIX
-    - nodejs_version: "7.4.0"
+    - nodejs_version: "8.11.3"
```
glennsarti: For reference - As of right now, the VSCode tests use Node 8.9.1 (https://vscode.visualstudio.com/VSCode/_build/results?buildId=4595&view=logs) so 8.11.3 should be fine.

Issue #314, opened 2018-07-17
Update VSCode engine to 1.27


jpogran:
We have left the minimum VSCode version needed at 1.11 since almost inception of the project. It's time to require a more recent version, as the [June 2018 update](https://code.visualstudio.com/updates/v1_25) adds some interesting possibilities for us

- [Outline view](https://code.visualstudio.com/updates/v1_25#_outline-view)
- [Document symbols](https://code.visualstudio.com/updates/v1_25#_document-symbols)
- [Custom tree views](https://code.visualstudio.com/updates/v1_25#_custom-views)
- [Language Server Protocol Inspector](https://code.visualstudio.com/updates/v1_25#_language-server-protocol-inspector)
- [Improved Quick Pick](https://code.visualstudio.com/updates/v1_25#_quickinput-api) (preview feature currently)

This issue may spawn other issues to update components due to the engine version increase.

PR #352, opened 2018-08-16
Revert "(GH-314) Update VSCode engine to 1.25"

glennsarti:
This reverts commit 9a2308019217df164e48707e73580fce5223a8d3.


jpogran:
This is done to allow a small bug-fix release that will not exclude users using older VSCode versions

Issue #366, opened 2018-08-24
Use the new Debug Configuration Provider


glennsarti:
The [Debug Extension API](https://code.visualstudio.com/docs/extensionAPI/api-debugging#_debug-extension-api) added a DebugConfigurationProvider feature which allowed for more granular control of how to launch a Debug Adapter.

We should use this and reduce the "smartness" of the Debug Adapter, which is basically a proxy.

PR #367, opened 2018-08-24
(GH-366) Refactor Debug Adapter

glennsarti:
This commit adds an implementation of a DebugConfigurationProvider;

* The adapterExecutableCommand settings is created, and a command registered
  which will return the command and arguments to start the Debug Adapter.  In
  this case the arguments will include the ruby environment information

* The provideDebugConfigurations method just returns the same debug
  configuration  as in package.json

* Modified the DebugAdapater.ts file to use the newly injected ruby information.
  This means the debug adapter no longer needs to share the code from the
  extension as itself.

* However an issue was found in the DebugAdapter, if it exited without sending
  the appropriate terminate method, VS Code would get in a weird state which
  required it to be reloaded to start debugging again.  To stop this, if the
  ruby based adapter is not spawned, a Null Debug Adapter will be started on
  STDIO which throws and error and terminates gracefully, thus VS Code does not
  get confused.

* Updated the default debug configuration to have noop = true by default

---

- [ ] Update README
- [ ] Consider name change

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,13 +173,17 @@ The Puppet extension is able to debug the compilation of a Puppet manifest; much
 * Call stack
 * Variables, but only at the top stack frame
 * Limited interactive debug console.  For example, you can assign a variable a value, but just as in regular Puppet you can't change its value later
-* Step In, Out, Over
-
-You may be presented with a Launch Configuration on first use. Please see the [VSCode Debugging link](https://code.visualstudio.com/docs/editor/debugging) for instructions on how to set this up.
+* Step In, Out, and Over
 
 The debugging features in the extension are based on the excellent ideas in [puppet-debugger](https://www.puppet-debugger.com/) by [Corey Osman](https://github.com/nwops).
 
-Settings:
+#### Configuring the debug session
+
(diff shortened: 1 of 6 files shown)


jpogran:
1. Checked out PR, packaged VSIX and installed.
2. Used example node graph PP file.
3. Hit debug (F5), nothing.
4. Hit Launch Settings icon, nothing. Hit drop down, add configuration. Searched for Puppet. Found 'Puppet Apply current file'. Was confused by name, allowed it to create entry.
5. Went back to PP file, hit F5, immediate 'Debug adapter process has terminated unexpectedly'.
6. Enabled "puppet.languageServer.debugFilePath", tried again. nothing logged to file, same error.

glennsarti:
Potential configuration name change:
* `Puppet: Launch current manifest`

jpogran:
@glennsarti you were correct, I was bitten by extension autoupdate.  This works correctly and fast!

We still need to work on the readme, so leaving that there.

jpogran:
I would be happy with 'Puppet Debug current file'

PR #368, opened 2018-08-27
(maint) Minor fixes

glennsarti:
The PuppeConfig file is not used anywhere.  This commit removes this redundant
file.

---

Previously verbose was the most verbose as the order of the enum is important.
This commit moves Debug to the top of the list which makes it the most verbose
level.


src/logging.ts
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 export enum LogLevel {
-  Verbose,
   Debug,
+  Verbose,
```
jpogran: Is changing how logs are written a minor thing?

glennsarti: IMO, in this case it is.  I'm not changing how logs are written per se, I'm fixing a bug in the filtering so it's more inline with what someone would expect.

Debug logging should create more "Stuff" than Verbose logging.

jpogran: I meant that you're fixing a bug, that's not a maint commit. It changes what is logged. If you had it verbose before to get all the logs (like me), this changes behavior.

PR #370, opened 2018-08-27
(GH-373) Refactor node graph into a feature

glennsarti:
Previously the node graph feature and provider was strewn across multple files.
This commit instead uses a Feature -> Provider pattern to supply the node
graph text.  Functionally there is no difference for the user.


src/feature/NodeGraphFeature.ts
```diff
@@ -0,0 +1,183 @@
+'use strict';
```
jpogran: Git nit: could this have been a file rename instead of a delete? Lost history...etc

src/feature/NodeGraphFeature.ts
```diff
@@ -0,0 +1,183 @@
...
+      this.waiting = true;
+      setTimeout(() => {
+        this.waiting = false;
+        this.onDidChangeEvent.fire(uri);
+      }, 300);
+    }
+  }
+}
+
+export class NodeGraphFeature implements IFeature {
```
jpogran: Is it a typescript thing to put multiple classes in one file? While this is all one 'feature', does it need to be all in one file?

src/feature/NodeGraphFeature.ts
```diff
@@ -0,0 +1,183 @@
...
+    connectionManager: IConnectionManager,
+    logger: ILogger,
+    context: vscode.ExtensionContext
+  ) {
+    context.subscriptions.push(vscode.commands.registerCommand(PuppetNodeGraphToTheSideCommandId,
+      uri => this.showNodeGraph(uri, true))
+    );
+    logger.debug("Registered " + PuppetNodeGraphToTheSideCommandId + " command");
+
+    this.provider = new NodeGraphContentProvider(connectionManager);
```
jpogran: Would it be better to pass in a `NodeGraphContentProvider` instead of an `IConnectionManager`, since the `ConnectionManager` object is only used by the `NodeGraphContentProvider`? This removes one reference, and allows us to change `ConnectionManager` independent of this class.

Or, does the array implementation of extensionFeatures in the main activation method not allow us to have dependencies between 'features'?

src/feature/NodeGraphFeature.ts
glennsarti: Rebased and fixed.

src/feature/NodeGraphFeature.ts
glennsarti: I forgot to remove the export from the provider.  From a TS point of view, only the feature class is exported.  so one class - one file.

src/feature/NodeGraphFeature.ts
glennsarti: Ideally I only wanted to pass in the LanguageClient object, however that needs a refactor of the connection manager which I don't want to do in this PR.

> Or, does the array implementation of extensionFeatures in the main activation method not allow us to have dependencies between 'features'?

Only becuase I've done it as a single line assignment.  There's nothing wrong with appending to the array on a per feature basis.

jpogran:
LGTM now

PR #371, opened 2018-09-04
(maint) Remove errant console.log

glennsarti:
Remove errant console.log


jpogran:
Doesn't removing this remove the output in the Debug console when launching the extensionhost debugger?

glennsarti:
Kind of....it should probably be going through the logger not console.log

glennsarti:
Fixed.  Now running through the logger.

Issue #373, opened 2018-09-04
Refactor code and group features and providers together


glennsarti:
**As a** Developer

**I would like to** group extension features into a singular file

**so that I** I don't need to have to modify multiple files when adding features

For example,
Puppet command text + Feature + register command etc.. Ideally it should just be a single feature file e.g.
 PuppetNodeGraph or PuppetResourceCommand.

- [x] Node Graph
- [x] Puppet Resource Command
- [x] Format Document
- [x] PDK NewClass Command
- [x] PDK NewModule Command
- [x] PDK NewTask Command
- [x] PDK Test Command
- [x] PDK Validate Command

glennsarti:
Raised #370 for the node graph

glennsarti:
PR #374 is for the Format Document provider.

glennsarti:
@jpogran I think this can be closed now.

jpogran:
![yes](https://media.giphy.com/media/3ohhweiVB36rAlqVCE/giphy.gif)

PR #374, opened 2018-09-17
(GH-373) Feature-ify Format Document provider

glennsarti:
Previously the format document feature and provider was strewn across multiple
files. This commit instead uses a Feature -> Provider pattern to supply the text
edits.  Functionally there is no difference for the user.


src/feature/FormatDocumentFeature.ts
```diff
@@ -0,0 +1,71 @@
...
+class FormatDocumentProvider {
+  private connectionManager: IConnectionManager = undefined;
+
+  constructor(connectionManager:IConnectionManager) {
+    this.connectionManager = connectionManager;
+  }
+
+  public formatTextEdits(document: vscode.TextDocument, options:vscode.FormattingOptions) : Thenable<vscode.TextEdit[]> {
+    if (this.connectionManager.status !== ConnectionStatus.Running ) {
+      vscode.window.showInformationMessage("Formatting Puppet files is not available as the Language Server is not ready");
```
glennsarti: I wonder if this is a good time to change the language of this error?

jpogran: I think for right now, adding 'Please wait and try again. The Puppet extension is still loading...' is good

glennsarti:
Text changed.

PR #377, opened 2018-09-20
(GH-373) Refactor PDKCommands to features

jpogran:
This converts all of the PDK commands to use the Feature infrastructure.

PR #378, opened 2018-09-20
(GH-373) Convert Puppet Resource command to Feature

jpogran:
This PR converts the Puppet Resource command to use the Feature infrastructure


jpogran:
Updated to not rely on #377

Issue #375, opened 2018-09-19
Support starting Language Server in Docker container


jpogran:
This issue is a result of spike issue #364. The conclusion of that ticket should be the purpose of this one

PR #387, opened 2018-10-02
(WIP)(GH-375) Puppet Editor Services in a Container

michaeltlombardi:
Ran some experiments, hooked up some plumbing for both docker commands and
settings. Cannot get the extension to start though when using the docker
settings, need to do more troubleshooting and investigating.

#375

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -31,6 +31,7 @@ export class ConnectionManager implements IConnectionManager {
   private languageServerProcess: cp.ChildProcess;
   private puppetLanguageClient: PuppetLanguageClient;
   private timeSpent:number;
+  private dockerName: string = undefined;
 
   constructor(
     context: vscode.ExtensionContext,
@@ -98,6 +99,9 @@ export class ConnectionManager implements IConnectionManager {
       this.languageServerClient.stop();
     }
 
+    // Kill the docker container
+    this.stopLanguageServerDockerProcess();
+
     // The language server process we spawn will close once the
     // client disconnects.  No need to forcibly kill the process here. Also the language
     // client will try and send a shutdown event, which will throw errors if the language
@@ -145,7 +149,7 @@ export class ConnectionManager implements IConnectionManager {
 
           // If the language client isn't already running and it's sent the trigger text, start up a client
           if (this.languageServerClient === undefined && /LANGUAGE SERVER RUNNING/.test(data.toString())) {
-            if(this.connectionConfiguration.port){            
+            if(this.connectionConfiguration.port){
               this.languageServerClient = this.createLanguageClient();
             }else{
               var p = data.toString().match(/LANGUAGE SERVER RUNNING.*:(\d+)/);
@@ -156,6 +160,17 @@ export class ConnectionManager implements IConnectionManager {
           }
         });
         break;
+        case ProtocolType.DOCKER:
+          this.languageServerProcess.stdout.on('data', data => {
+            this.logger.debug('OUTPUT: ' + data.toString());
+
+            // If the language client isn't already running and it's sent the trigger text, start up a client
+            if (this.languageServerClient === undefined && /LANGUAGE SERVER RUNNING/.test(data.toString())) {
+              this.languageServerClient = this.createLanguageClient();
+              this.extensionContext.subscriptions.push(this.languageServerClient.start());
+            }
+          });
+          break;
       case ProtocolType.STDIO:
         this.logger.debug('Starting STDIO client: ');
         this.languageServerClient = this.createLanguageClient();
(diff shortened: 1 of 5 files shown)


michaeltlombardi:
Update! So, got 90% of the plumbing to work... except it won't start the container. Considering switching to using `docker run --rm -it` and connecting via stdio instead of TCP.

Once we have a POC in place, will need to completely rework, I think (it's v messy).

We can come back around and do some plumbing such that there's a `useDocker` boolean setting and some subsettings allowing users to specify language server image and tag.

We _should_ (?) be able to allow users to specify whether they want to connect to the container via TCP or stdio.

More to follow!

michaeltlombardi:
We did it!

michaeltlombardi:
So, latest push has the code "working", wherein it spins up the language server inside a docker container, but absolutely everything is hardcoded and awful.

Tomorrow:

Plumbing time

jpogran:
yes

michaeltlombardi:
Update: Exactly as many explosions and horrors as predicted. Cleaned out the old code and.... now it yells immediately, closes the connection in rapid succession, and ducks out.

Luckily, the container still isn't disposed of, so when you reload it "just works".....

More to follow, more to debug.

michaeltlombardi:
Okay, so! We have the plumbing set up such that we can start the container reliably without injecting delays. However! We still can't cleanly _stop_ the container.... for reasons? Trying to figure out if I can call an appropriate kill to the process which will respect the `--rm` flag, removing the container.

Will go back through and also enable the port/tag information to be set/used from the config instead of hard-coding.

The primary blocker right now is on destroying the container at the end of the run.

michaeltlombardi:
So, I've been experimenting around with disposing of the process running the container...

(\° -°)\ ┬──┬

My entire approach to stopping the containers [wasn't going to work at all because I failed a reading comprehension check](https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/).

(╯°□°）╯︵ ┻━┻ 

TLDR: I can't manage the container by managing the process because sending a kill to it doesn't, somehow, kill the the attached container. So I'll need to think of another option - but, spinning up the container the way we do now lets us reuse all the other plumbing... and doesn't tell us the container's name or id, so I can't use that to later kill it.

Further investigation shows that the way we have the plumbing _now_ (using stdout from the process to determine when to try to connect with the client) won't be reusable, meaning we'll need to write more stuff there.

A terrible stopgap would be to kill all running containers with the specified repo/name/tag combo... which I'm not fond of, but I also don't see good reasons for folks to be running lots of them outside of VSCode? Flip side, I don't yet know if one container can serve 2+ sessions... meaning closing it down would gobble up the others.

michaeltlombardi:
Hurray! We have a bunch more stuff working now!

- [x] Randomized Ports
- [x] Randomized Name
- [x] Cleanup on close
- [x] Reliable spinup

Just going to update the config portion so there's less hardcoding and see if we can clean some stuff up - will need to test and see if some of the 'safety' code I just copied is necessary.

src/settings.ts
```diff
@@ -131,6 +132,11 @@ export function settingsFromWorkspace(): ISettings {
     enable: true,
   };
 
+  const defaultDockerSettings: IEditorServiceDockerSettings = {
+    imageName: "linguapupuli/puppet-langauge-server",
```
michaeltlombardi: This includes the misspelled image, will probably want to fix that before we ship.

src/interfaces.ts
```diff
@@ -20,6 +20,7 @@ export enum ProtocolType {
   UNKNOWN = '<unknown>',
   STDIO = 'stdio',
   TCP = 'tcp',
+  DOCKER = 'docker'
```
michaeltlombardi: This may not be the _right_ way to do things but it enabled us to do some easy plumbing. Could probably be reworked?

src/connection.ts
```diff
@@ -182,6 +197,74 @@ export class ConnectionManager implements IConnectionManager {
...
+  // TODO: Should this return anything? /shrug
+  private stopLanguageServerDockerProcess(): void {
+    let logPrefix: string = '[stopLanguageServerProcessAsContainer] ';
+    let cmd: string = 'docker.exe';
+    let args: Array<string> = [
+      "rm",
+      '--force',
+      this.dockerName,
+    ];
+    //TODO: Is this necessary?
```
michaeltlombardi: Wasn't sure if the spawn_options are actually needed. If not, we can clean out some duplicate code.

src/connection.ts
```diff
@@ -182,6 +197,74 @@ export class ConnectionManager implements IConnectionManager {
...
+      '--rm',
+      '-i',
+      '-P',
+      '--name',
+      this.dockerName,
+      `${this.connectionConfiguration.dockerImageName}:${this.connectionConfiguration.dockerImageTag}`
+    ];
+    console.log(vscode.CompletionItemKind.Class);
+    // Is this windows only? When run on windows PATH is just the ruby paths, Path has everything else.
+    options.env.PATH = process.env.PATH;
```
michaeltlombardi: This had some strangeness - on my local machine PATH only included the ruby stuff, not sure if this was intentional - but we overwrite back to the process environment variable to move forward for docker.

There's no generalized "is docker there" test yet, not sure where to put that - should probably dump out relatively gracefully if it can't find docker in the `PATH`.

src/connection.ts
```diff
@@ -182,6 +197,74 @@ export class ConnectionManager implements IConnectionManager {
...
+    // We want to retrieve that port for use in the TCP connection.
+    let regex = /:(\d+)$/m;
+    return Number(regex.exec(proc.stdout.toString())[1]);
+  }
+
+  public startLanguageServerDockerProcess( options: cp.SpawnOptions, callback: Function) {
+    let logPrefix: string = '[startLanguageServerProcessAsContainer] ';
+    this.dockerName = this.createUniqueDockerName();
+    // TODO: get settings to specify image name and tag
+    let cmd: string = 'docker.exe';
```
michaeltlombardi: This should probably be gates for *nix/windows.

src/connection.ts
```diff
@@ -182,6 +197,74 @@ export class ConnectionManager implements IConnectionManager {
...
 
+  private createUniqueDockerName() {
+    return 'puppet-vscode-xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+      const r = Math.random() * 16 | 0, v = c === 'x' ? r : ( r & 0x3 | 0x8 );
+      return v.toString(16);
+    });
+  }
+
+  private getDockerPort(name: string) {
+    let cmd: string = 'docker.exe';
```
michaeltlombardi: This should probably be gated for *nix/windows.

michaeltlombardi:
Fully plumbed!

So, some considerations:

- This has only be checked on Windows, needs more love for OSX and Linux.
- There's some duplicate code that could probably be cleaned up (depending on whether it's actually necessary).
- The only configurable knobs right now are image name and tag.
- Connections are always over TCP; might want to rewrite this to be a startup option (docker or local process) and let the user further select stdio vs TCP?
- We'll need a rebuild of the language server container prior to release with a corrected name (`language` instead of `langauge`).

Learned a bunch of stuff along the way though!

michaeltlombardi:
Checked in with the pupperware folks, looks like the [puppet-agent image matrix is being actively maintained](https://github.com/puppetlabs/puppet-agent/pull/1543), but slimmed down - for now, to just ubuntu and alpine - but that should be merged and available soon, so we should be able to work on our image stuff.

michaeltlombardi:
[Puppet Agent Alpine Image is live and updated](https://hub.docker.com/r/puppet/puppet-agent-alpine/tags/), will need to rebuild our image on `6.0.2`.

michaeltlombardi:
Bounced to force CI.

jpogran:
With the refactor done in #425 this needs to be rebased and reworked to use the new ConnectionHandler abstract class.

jpogran:
This awesome work lead to PR #450 being implemented. We salute this work, and close it with honor

PR #389, opened 2018-10-07
(GH-373) Convert Commands to Features

jpogran:
This PR converts all the remaining commands to the new Feature infrastructure.

The last commit stretches the definition of maint, as it extracts code to methods and does alot of re-organizing but results in a cleaner extension.ts file.


glennsarti:
- [x] Puppet Resource
- [x] Status Bar
- [x] Session popup menu (Shows and does what's expected)
- [ ] PDK commands

PR #393, opened 2018-10-10
(GH-399) Update status bar loading UI

### LINKED ISSUE #399: Update Puppet loading UI to be more user friendly
Previously the UI when puppet was loading implied that the extension wasn't running, however this isn't true. 

* The loading UI should indicate that the extension is somewhat operational i.e. As soon as the server is running and connected, it will have hover support BUT it may not return useful information yet.

* The loading UI should make it easier to discover what's available and what is not.

glennsarti:
Previously the UI when puppet was loading implied that the extension wasn't
running, however this isn't true. This commit changes the loading and
connection status so that we know when the Language Server is running, but
partially loaded vs fully loaded.

Fixes #399

### DIFF
--- a/src/PuppetLanguageClient.ts
+++ b/src/PuppetLanguageClient.ts
@@ -42,20 +42,20 @@ export class PuppetLanguageClient {
     this.languageServerClient.onReady().then(
       () => {
         logger.debug('Language server client started, setting puppet version');
-        this.setConnectionStatus('Loading Puppet', ConnectionStatus.Starting);
+        this.setConnectionStatus('Loading Puppet', ConnectionStatus.Starting, '');
         this.queryLanguageServerStatus();
       },
       reason => {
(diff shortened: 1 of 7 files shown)


glennsarti:
UI new look

statusbar-change

glennsarti:
I'm not 100% sure on the gear icon when loading, but the rest I'm really happy with.

jpogran:
I'm thinking that we should include an info message window that pops a list of things available at start that the user does not have to wait for a full puppet load, or explains that the extension is ready to use, just wait a little longer for X, Y, and Z.

glennsarti:
I'm still hesitant to display an info message window.  They are costly/heavy handed to use because they interrupt a user.

Possible other options;
* Show the information in the tool tip while load
* Or when you click on the status bar there's a "What does this mean?" command which shows the window?

glennsarti:
Ok new version uses Tool tips to display more information

statusbar-change

Issue #411, opened 2018-10-24
Remove internal information from the README


glennsarti:
In the README we list the symbols we support, but we also list the internal Puppet Pops type.  Users don't actually know what that is, indeed most people don't care.

I think it should be removed, and just list the common names e.g.

`Supported symbols are Class, Defined Types, Attributes and Variables`

Instead of;

Supported symbols:

Symbol | Puppet Class
 -- | --
Class |  `Puppet::Pops::Model::HostClassDefinition`
Defined Type | `Puppet::Pops::Model::ResourceTypeDefinition`
Attribute | `Puppet::Pops::Model::AttributeOperation`
Variable | `Puppet::Pops::Model::AssignmentExpression`

glennsarti:
Interested in your opinion @michaeltlombardi

michaeltlombardi:
I concur with removing the puppet class information, only the symbols really matter to the end user.

jpogran:
Complete

PR #414, opened 2018-10-29
(GH-411) Remove class name for symbols from README

jpogran:
This commit removes the Puppet class names for the document symbols that
are supported for better readability.


jpogran:
Bitten by newlines. Please standby

PR #435, opened 2018-11-29
(GH-434) Update to 0.16.0 puppet-editor-services

jpogran:
This commit pins to 0.16.0 version of puppet-editor-services.


glennsarti:
Blocked on 0.16.0 release

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how
 
 ## Unreleased
 
+### Changed
```
jpogran: @michaeltlombardi keep me honest with the changlog format

michaeltlombardi: These all work for the `Changed` section, but the changelog here doesn't specify the new features or bug fixes - this is user facing, so anything the end-user of the editor services would notice (particularly new features) should probably be included here, even if the actual work happened in a related project we consume.

I'm not sure about this long term, but right now this extension drives the development of those other independent projects.

README.md
```diff
@@ -163,6 +164,14 @@ As part of IntelliSense and Snippets, you can quickly create templates to automa
 
 Linting is automatically applied to the document as you edit, without having to save the file first. The validation uses [puppet-lint](https://github.com/rodjek/puppet-lint) to validate the Puppet syntax in all open puppet files. Errors and warnings appear in the Problems window in VSCode.
 
+### Live Workspace Intellisense
+
+If you are editing a module you will now get live intellisense of the [Puppet Custom Types](https://puppet.com/docs/puppet/latest/custom_types.html), [Functions](https://puppet.com/docs/puppet/latest/lang_write_functions_in_puppet.html), [Classes](https://puppet.com/docs/puppet/latest/lang_classes.html) and [Defined Types](https://puppet.com/docs/puppet/latest/lang_defined_types.html). For example if you are editing the [puppetlabs-apache module](https://github.com/puppetlabs/puppetlabs-apache) you will be able to get auto-complete for resources like `apache::vhost` or `apache::mod::php`.
+
+Support for [Control Repositories](https://puppet.com/docs/pe/latest/control_repo.html) is _coming soon!_
```
michaeltlombardi: 😍

CHANGELOG.md
jpogran: I first tried this out by copying the changelog from puppet-editor-services here, but that didn't work so well. Since we use markdown links, it looked just like puppet-vscode issues but the puppet-editor-services numbering was lower than puppet-vscode so the numbers looked off. I then thought to add puppet-editor-services somewhere in the display part, but that meant I would have to change everything to match or else it would look weird.

So, I put up links to the changelog in each line instead that referred to a different project

PR #438, opened 2018-12-03
(maint) Add uneeded files to vscodeignore

jpogran:
This adds CI files and github templates to the vscodeignore file.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -11,3 +11,6 @@ logs/**
 sessions/**
 gulpfile.js
 docs
+.github
+.travis.yml
+appveyor.yml


.vscodeignore
```diff
@@ -11,3 +11,6 @@ logs/**
 sessions/**
 gulpfile.js
 docs
+.github
```
glennsarti: SHould this be a directory or doesn't matter?

jpogran: Does not appear to matter to vscode, as it kept the directory out. I'll change it anyway

glennsarti:
Bouncing the PR to force Appveyor to redo

PR #442, opened 2018-12-05
(GH-335) Add option to disable language server

### LINKED ISSUE #335: Allow disabling of language server
I get this error whenever I open a `.pp`:

```
Could not find a valid Puppet installation at '/opt/puppetlabs/puppet'. While syntax highlighting and grammar detection will still work, intellisense and other advanced features will not.
```

I'm happy just using the syntax highlighting and grammar detection (as stated), and not having VS Code interact with a language server at all.

I seem to recall that there used to be an option to turn it off - but can no longer find it.

jpogran:
This PR surfaces an option to disable loading the Puppet Language
Server, while still allowing the features that do not require a language
server. The first time this option is set, the user will be presented
with a message explainging what they turned off with an option to not
show this message again. The message won't be shown next time if the
user presses the button.

Fixes #335

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -19,7 +19,7 @@ import { OutputChannelLogger } from './logging/outputchannel';
 import { PuppetCommandStrings } from './messages';
 import { PuppetStatusBar } from './PuppetStatusBar';
 import { ISettings, legacySettings, settingsFromWorkspace } from './settings';
-import { Reporter } from './telemetry/telemetry';
+import { Reporter, reporter } from './telemetry/telemetry';
 
 const langID = 'puppet'; // don't change this
 let extContext: vscode.ExtensionContext;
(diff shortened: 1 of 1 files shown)


jpogran:
@glennsarti @michaeltlombardi this works to stop the extension from ever attempting to load the language server and doesn't affect closing the editor since features array is null and the handler object is never initialized. However there is one feature that (currently) doesn't need the language server: pdk commands. I know we are planning on moving those to the langserver, so is it worth it to complicate the code enough to load the PDK features even if we don't load the langserver?

michaeltlombardi:
In my opinion: No, it's not. I don't know how often folks will want to use the extension without language support, so I'd hold off until we get (more?) user feedback asking for it. Especially with the complexity cost and the eventual moving of PDK stuff to the language server.

michaeltlombardi:
Actually read the user issue (should've done that first): I now vote 'hold off' rather than 'do not implement', assuming the PDK-> Language Server move is in the next 2-3 releases.

jpogran:
Alright, moving this off of WIP. As its implemented now, it does not provide anything other than syntax-highlighting and minimal autocomplete.

PR #457, opened 2019-01-07
(Maint) Update vendoring for Editor Syntax

glennsarti:
Previously only the editor services was vendored however we need to support multiple resouces.
This PR refactors the gulp file vendoring and adds the Editor Syntax resource.


vendoring.json
```diff
@@ -0,0 +1,8 @@
+{
```
jpogran: Not sure on the name for this file. I like the `editor-` from the previous incarnation. How about `editor-components.json`? It keeps it specific to us, and (to me) does not imply a general dependency file. I may be over thinking this

jpogran:
Great work!

glennsarti:
Renamed.

Issue #454, opened 2019-01-04
PDK 1.8.0 does not work


jpogran:
The vendored ruby versions have changed in pdk 1.8.0 and the extension can no longer find the correct paths

PR #460, opened 2019-01-08
(GH-454) Fix PDK Ruby file path

jpogran:
This commit updates the default Ruby version from 2.4.0 to 2.5.1 used
when using PDK as the source for the Puppet Language Server. Ruby 2.5.1
is in PDK 1.7.1 and PDK 1.8.0 so it is safe to use as default until we
can land a feature to allow the user to pick between available PDK ruby
versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@
     - [Puppet Resource](#puppet-resource)
     - [Puppet Node Graph](#puppet-node-graph)
   - [Puppet Development Kit Support](#puppet-development-kit-support)
+    - [PDK Supported Versions](#pdk-supported-versions)
     - [PDK New Module](#pdk-new-module)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -213,6 +214,10 @@ You can use the [Puppet Development Kit](https://puppet.com/blog/develop-modules
 
 > Note: `pdk convert` is not available in the command palette as it is a complicated command that requires user input to succeed. It is better to use it from the builtin terminal.
 
+#### PDK Supported Versions
+
+The Puppet VSCode Extension supports the current PDK version, and one older version. This is currently 18.0 and 1.7.1.
```
glennsarti: Typo   `1.8.0`

jpogran:
@glennsarti typo fixed

PR #468, opened 2019-01-19
(maint) Update telemetry

jpogran:


glennsarti:
The "hidden" commands thing looks like we need to use `editorLangId`  instead of  `resourceLangId`

PR #470, opened 2019-01-24
(GH-469) Restore puppet palatte commands

### LINKED ISSUE #469: Puppet Resource and Puppet Node Graph commands hidden
It appears that due to the addition of the puppetfile language to the extension, the `puppet resource` and the `puppet node graph` commands no longer appear in the command palate. 

This seems to be due to the following lines:

https://github.com/lingua-pupuli/puppet-vscode/blob/15aa7718534edecad5b8e5b34edba048c339c562/package.json#L190

We apparently need to be using `editorLangId`: https://github.com/Microsoft/vscode/issues/26873

jpogran:
When adding support for `Puppetfile` as a language it changed the
behavior of `resourceLangId` with contributed commands. This hid any
commands with `when` clauses that use `resourceLangId` to determine
their availability. Changing the `when` clauses to use `editorLangId`
has restored these commands in both the command palatte and in the
editor menus.

Fixes #469

### DIFF
--- a/package.json
+++ b/package.json
@@ -187,11 +187,11 @@
         },
         {
           "command": "extension.puppetResource",
-          "when": "resourceLangId == 'puppet'"
+          "when": "editorLangId == 'puppet'"
         },
         {
           "command": "extension.puppetShowNodeGraphToSide",
-          "when": "resourceLangId == 'puppet'"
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
Tested using artifact from appveyor, commands are discoverable and function as expected.

PR #473, opened 2019-01-25
(release) 0.16.0

jpogran:
Pending #470


jpogran:
re-pushed with correct date

PR #482, opened 2019-02-11
(GH-480) Automatically find PDK Ruby Version

### LINKED ISSUE #480: Language Server doesn't start with PDK 1.9.0 due to new Ruby version
### What Versions are you running?

OS Version: Win10
VSCode Version: 1.30.2
Puppet Extension Version: 0.16.0
Puppet Agent Version: N/A
PDK Version (if applicable): 1.9.0

### What You Are Seeing?

On opening a `.pp` file, in the Puppet output panel:

```[Error - 1:00:25 PM] Starting client failed
Launching server using command C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.5.1\bin\ruby failed.
[Error - 1:00:25 PM] Launching server using command C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.5.1\bin\ruby failed.
```

### How Did You Get This To Happen? (Steps to Reproduce)

PDK 1.9.0 includes Ruby 2.5.3 rather than 2.5.1

jpogran:
Fixes #480

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -2,6 +2,7 @@
 
 import * as vscode from 'vscode';
 import * as path from 'path';
+import * as fs from 'fs';
 
 import { IConnectionConfiguration, ConnectionType, ProtocolType, PuppetInstallType } from './interfaces';
 import { PathResolver } from './configuration/pathResolver';
@@ -13,8 +14,6 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
   public timeout: number;
   public debugFilePath: string;
   private settings: ISettings;
-  private gemRubyVersionDir: string;
-  private rubyVersionDir: string;
 
   constructor() {
     this.settings = settingsFromWorkspace();
@@ -25,9 +24,6 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
     this.debugFilePath = this.settings.editorService.debugFilePath;
 
     this._puppetInstallType = this.settings.installType;
-
-    this.rubyVersionDir    = '2.5.1';
-    this.gemRubyVersionDir = '2.5.0';
   }
 
   private _puppetInstallType: PuppetInstallType;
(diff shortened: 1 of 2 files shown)


src/configuration.ts
```diff
@@ -245,7 +246,21 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
   get pdkGemVerDir(): string {
     var rootDir = path.join(this.pdkRubyDir, 'lib', 'ruby', 'gems');
 
-    return path.join(rootDir, this.gemRubyVersionDir);;
+    return this.findFirstDirectory(rootDir);
+  }
+
+  private findFirstDirectory(rootDir: string) {
+    var files = fs.readdirSync(rootDir);
+    let result = files.sort( (a, b) => a.localeCompare(b, undefined, { numeric:true }) ).reverse()[0];
```
jpogran: This is rather naive, and kinda what got us hardcoding things in the first place. However, the logic works in that the versions are sorted in descending order this time around instead of assuming the one we want is the last in the list. The sorted 'first' is our latest, and the one we should use both for ruby version as well as for the cached version.

This sort can fail if there are weird version numbers that don't match semver, but ruby generally follows these and Puppet doesn't vendor interim versions so PDK doesn't either.

glennsarti:
Is this still WIP? @jpogran

jpogran:
Looking for validation on my naive sorting approach before I took off the WIP

src/configuration.ts
glennsarti: I agree this is naive but it may be exactly what we need right now (as opposed to hardcode).  We should definitely prioritise fixing this properly post release.

src/configuration.ts
```diff
@@ -245,7 +246,21 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
...
+    return path.join(rootDir, result);
+  }
+
+  private replaceSlashes(lib: string): string {
```
glennsarti: nit - Could probably use a better variable name than `lib`

src/configuration.ts
```diff
@@ -207,35 +208,35 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
...
-      lib = lib.replace(/\\/g, '/');
-    }
+    lib = this.replaceSlashes(lib);
     return lib;
   }
 
   get pdkRubyVerDir(): string {
     var rootDir = path.join(this.puppetBaseDir, 'private', 'puppet', 'ruby');
 
-    return path.join(rootDir, this.gemRubyVersionDir);
```
glennsarti: Don't forget to delete the references to `gemRubyVersionDir` and `rubyVersionDir` in the class definition and constructor.

src/configuration.ts
jpogran: Was artifact from extract method refactor. Changed now :)

src/configuration.ts
jpogran: Thanks for the reminder

PR #486, opened 2019-02-16
(GH-489) Dynamically detect PDK information

glennsarti:
Builds on #490 

---

Previously all of the logic used to detect PDK information was contatined within
the configuration class.  This commit adds a new PDK Resolver file which
contains classes to detect all of the available ruby instances within a PDK
installation, and detect the available Puppet gem versions for each ruby
instance.

This commit modifies the configuration class to use the new PDK Resolver.
This implements the similar configuration feature of choosing the latest
PDK ruby instance to use for the Puppet Editor Services.

---

- [x] Detect ruby versions
- [x] Detect puppet versions inside ruby versions
- [x] Tests 😢 Needed. - TBA
- [x] ~~Surface this to users?~~ Different ticket/project #489
- [x] ~~Add to right click menu on Puppet version to switch it?~~  Different ticket/project (We'd need to make sure PES honors this) #489

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -5,6 +5,7 @@ import * as fs from 'fs';
 
 import { PathResolver } from './configuration/pathResolver';
 import { ISettings, ConnectionType, ProtocolType, PuppetInstallType } from './settings';
+import * as pdk from './configuration/pdkResolver';
 
 /** Creates an Aggregate Configuration based on the VSCode Workspace settings (ISettings) */
 export function CreateAggregrateConfiguration(settings:ISettings): IAggregateConfiguration {
@@ -63,9 +64,16 @@ export class AggregateConfiguration implements IAggregateConfiguration {
 
     const puppetBaseDir = this.calculatePuppetBaseDir(settings);
     const puppetDir = this.safeJoin(puppetBaseDir, 'puppet');
-    const pdkRubyDir = this.findFirstDirectory(this.safeJoin(puppetBaseDir, 'private', 'ruby'));
     const facterDir = this.safeJoin(puppetBaseDir, 'facter');
     const rubyDir = this.calculateRubyDir(puppetBaseDir);
+
+    let pdkInstance: pdk.IPDKRubyInstance = pdk.emptyPDKInstance();
+    if (settings.installType === PuppetInstallType.PDK) {
+      const result = pdk.pdkInstances(puppetBaseDir).latest;
+      // An undefined instance means that either PDK isn't installed or that
+      // the requested version doesn't exist.
+      if (result !== undefined) { pdkInstance = result; }
+    }
 
     this.ruby = {
       puppetBaseDir: puppetBaseDir,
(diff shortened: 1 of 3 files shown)


glennsarti:
The PDK resolver spits out an object array like:

```
------
[{rubyVersion: "2.1.9"
, rubyDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.1.9"
, rubyVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\puppet\ruby\2.1.0"
, gemVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.1.9\lib\ruby\gems\2.1.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.1.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.1.0"
, puppetVerions: "4.10.12,4.7.1,4.8.2,4.9.4"
, valid: "true"}

, {rubyVersion: "2.4.4"
, rubyDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.4.4"
, rubyVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\puppet\ruby\2.4.0"
, gemVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.4.4\lib\ruby\gems\2.4.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.4.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.4.0"
, puppetVerions: "5.0.1,5.1.0,5.2.0,5.3.7,5.4.0,5.5.6"
, valid: "true"}

, {rubyVersion: "2.5.1"
, rubyDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.5.1"
, rubyVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\puppet\ruby\2.5.0"
, gemVerDir: "C:\Program Files\Puppet Labs\DevelopmentKit\private\ruby\2.5.1\lib\ruby\gems\2.5.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.5.0"
, gemDir: "C:\Program Files\Puppet Labs\DevelopmentKit\share\cache\ruby\2.5.0"
, puppetVerions: "6.0.2"
, valid: "true"}
]
```

glennsarti:
@jpogran Looking for your thoughts on this approach.

src/configuration.ts
```diff
@@ -208,46 +210,36 @@ export class ConnectionConfiguration implements IConnectionConfiguration {
   }
 
   get pdkRubyVerDir(): string {
-    var rootDir = path.join(this.puppetBaseDir, 'private', 'puppet', 'ruby');
-
-    return this.findFirstDirectory(rootDir);
+    this.getPDKInstance();
```
jpogran: Instead of having to call `this.getPDKInstance()` each time for each property, can the contents of `getPDKInstance` be put inside the constructor?

jpogran:
As for the other bullets in the TODO, i think those are separate tickets

src/configuration.ts
glennsarti: IMO, I try to keep as much "magic" as possible out of object constructors.  Makes it easier to overload methods or change things.  Also makes testing easier.  I class detecting PDK instances as "magical"

I kept the original methods e.g. `pdkRubyVerDir` etc. to keep the amount of change to a minimum, but  we could solve this in one of two ways.

1. Remove these "legacy" methods and create a new method called `pdkInstance` that returns an IPDKInstance object. Basically handing off the repsonsibility to the object consumers

2. I could use a oneliner instead for all of the "legacy" methods.
`return this.getPDKInstance().rubyVerDir;`

glennsarti:
Yeah, I suspected they were a bit "scope-creep"-y.  I think integration tests may be in order though e.g.

Install the PDK and then create the PDKInstance object and make sure it's populated.  What do you think?

glennsarti:
For reference:  Loading a specific version of a gem without using bundler etc. (maybe?)

https://stackoverflow.com/questions/2693862/how-do-i-require-a-specific-version-of-a-ruby-gem

src/configuration.ts
jpogran: After discussion, I think this is fine as is now. If it should change, the next tickets to use this to allow user to change version should refine it.

jpogran:
Regarding testing, I think its problematic because we would need pdk installed in appveyor and travis to do that, and install times time out appveyor. I think we could add the tests as pending if mocha supports that, or as commented out , then once we get test harness matured we can do that

glennsarti:
Hrmm the travis error is weird.

glennsarti:
Travis failure is due to #492

Issue #489, opened 2019-03-01
Allow users to change puppet version used by the language server in the PDK


glennsarti:
**Summary of the new feature**

As part of #480 it will soon be possible to detect all of the available Puppet versions within PDK installations.  Now that we have that information we should surface that to users so they can:

1. Specify a version they want in the configuration file
2. Have a selection list so users can pick from the available list.

---
**As a** user

**I would like to** choose which Puppet Versions in VSCode

**so that I** I can use the most appropriate version for my environment.

glennsarti:
Work In Progress

- [ ] Refactor the configuration settings to make PDK information easier to inject - PR #490
- [ ] Actually enumerate ruby gem instances inside PDK when needed - PR TBA
- [ ] Add a configuration item to configure which Puppet version to use - PR TBA
- [ ] Add a menu item to pick a version and initiate an extension restart - PR TBA

glennsarti:
All PRs merged.  All ducks in a row.  This issue can now be closed.

PR #490, opened 2019-03-02
(GH-489) Refactor configration classes 

glennsarti:
Previously the workspace and editor configuration was spread over many classes
and interfaces.  This commit;

* Uses a single interface IAggregateConfiguration which aggregates the disparate
  information into a single, and importantly, static place

* Note that information inside IAggregateConfiguration should be considered
  static and derived from ISettings class used to create the configuration.
  Ideally the property values should not be overwritten

* The default settings have been split out into its own method to make testing
  easier.

* A helper class called safeJoin which can "join" paths which contain an
  undefined element

* Updated all of the other dependent classes to only pass through a single
  IAggregateConfiguration object in object constructors


glennsarti:
@jpogran Interested in your opinion here.

Does this seem like the right course of action; Create an AggregateConfiguration class, which is the aggregate of all our "static" configuration information:

- VSCode Workspace Settings
- Derived Ruby settings (including PDK bin stuff)
- Language Server Connection settings

jpogran:
One class to rule config settings seem to be the best option here for now. Naming is hard, so I'm fine with the name as is, can be changed later if it really becomes confusing. Only other comment is a nitpick, would have preffered to have it be several files inside a /configuration folder instead of one massive file, but that's an accepted pattern in typescript/javascript so I just need to deal

PR #491, opened 2019-03-07
(GH-489) Use alternate puppet version

glennsarti:
TODO

- [x] Update docs with GIF of the feature
- [ ] Test cross platform

---

This commit updates Puppet Editor Services to 0.19.0 as the puppet-version
command line argument feature is required.

---

Previously the status bar was spread of a few files.  This commit refactors the
Puppet status bar in the vscode into a single feature.  This will enable
changes to the status bar to be made easier in the future.

---

Previously when the PDK resolver was introduced it only listed the latest Puppet
version.  This commit:

* Adds a new property to the PDK resolver which lists all puppet versions
  available via the PDK.

* Adds a method to be able to search for a PDK instance for a particular Puppet
  version.

* Updates the TypeScript configuration to use the es2016 libraries so that the
  includes method can be used on array prototypes.

---

Previously Puppet Editor Services (PES) was left to use the latest puppet gem
version. However PES can now be passed a --puppet-version parameter to specify
which puppet version to use. This commit:

* Updates the extension settings to allow a user to set the requested version

* Updates the configuration to search for the specified puppet version and use
  the appropriate PDK instance (specifically Ruby version and Gem cache) for
  the ruby settings

* Passes this setting through into PES.

---

Now that the extension is able to query the available puppet versions and
specify a setting to actually use them, the User Interface needs to be updated
to easily helper users to change this.  This commit:

* Updates the Puppet Status Bar, so that if a user clicks on the Puppet version
  they are presented a list of available options, which the user can select
  from.

* Adds a new command which can be used to update arbitrary workspace settings.
  The version switching menu items call this configuration update feature to
  update the 'puppet.editorService.puppet.version' setting as per the user's
  request.

* The update configuration feature can optionally prompt the user to restart
  VSCode if the setting is known to require a restart.


src/feature/PuppetStatusBarFeature.ts
```diff
@@ -64,15 +71,39 @@ export class PuppetStatusBar {
...
     var menuItems: PuppetConnectionMenuItem[] = [];
 
     menuItems.push(
       new PuppetConnectionMenuItem(
         "Show Puppet Session Logs",
         () => { vscode.commands.executeCommand(PuppetCommandStrings.PuppetShowConnectionLogsCommandId); }),
     );
-  
+
+    if (this.config.ruby.pdkPuppetVersions !== undefined && this.config.ruby.pdkPuppetVersions.length > 0) {
```
jpogran: `this.config.ruby.pdkPuppetVersions` is 0 on the mac, but greater than 0 on windows

glennsarti: Found the problem and fixed it.  Adding tests now.

glennsarti:
Wohoo!!  working in Ubuntu 18.04 with PDK 1.9.1

Select version
image


... and restarting
image

src/feature/PuppetStatusBarFeature.ts
glennsarti: Fixed!!!

Issue #502, opened 2019-05-02
Switch default PuppetInstallType to PDK rather than the agent


albatrossflavour:
**Summary of the new feature**

As a user, I would like to be able to install the plugin and use it with the minimum possible steps.  Currently the PuppetInstallType is set to `agent`, which means you have to make a config update to change it to `PDK` unless you have an agent installed on the workstation.

Before PDK, this was pretty common to do, but now I just rely on having PDK there.

I spoke with @jpogran during the office hours session on 02/05/2019 and he suggested I raise a ticket for it.  I've got [a fork](https://github.com/albatrossflavour/puppet-vscode.git) that I'll raise a PR for shortly.

jpogran:
Thanks for the work @albatrossflavour! This is now merged in

PR #503, opened 2019-05-02
(GH-502) change the default install type to PDK

albatrossflavour:
https://github.com/lingua-pupuli/puppet-vscode/issues/502

Change the PuppetInstallType default to PDK

This will reduce the number of steps to getting the plugin working now that PDK is available.

Could not do testing as detailed in the contrib guide as there are no spec tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -396,7 +396,7 @@
         },
         "puppet.installType": {
           "type": "string",
-          "default": "agent",
+          "default": "pdk",
           "description": "The type of Puppet installation. Either the Puppet Development Kit (pdk) or the Puppet Agent (agent)",
           "enum": [
             "pdk",
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -163,7 +163,7 @@ export function DefaultWorkspaceSettings(): ISettings {
       enable: true
     },
     installDirectory: undefined,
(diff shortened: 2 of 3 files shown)


jpogran:
Thanks for the PR! Apologies on the delay in response, I was on medical leave for awhile.

We'll also need to update the `installType` line in the `package.json` to `PDK` as well. 

The tests are failing because they are testing to see that `agent` is the default, so we'll have to change those as well.

Don't worry about the contrib guide testing, thats for us to do.

glennsarti:
@jpogran Is this still WIP? or you want a review from me?

jpogran:
@glennsarti yep. Need you or @michaeltlombardi or @albatrossflavour to take it for a spin just to make sure it works as intended. I'm being over cautious.

jpogran:
The auto approach would be a separate ticket, this is still valid as setting what the default option will be. I think we should merge this as is with a ticket for the auto approach, both landing in the next release

glennsarti:
Can we check the analytics to see how many people haven't set the installType.   This still feels like it's going to break a lot of people, but the data should help with that.

jpogran:
We're going to merge this in as is, and then work on an auto detect feature

PR #510, opened 2019-05-15
(GH-505) PDK New Module Button Visibility Setting

### LINKED ISSUE #505: Unable to hide PDK New Module button
**As a user, I would like to** be able to hide the "PDK New Module" button on the top right corner of the VS Code window. There doesn't appear to be a setting to hide it, and since it's an infrequently used feature for me I would like to reduce visual noise.

Please let me know if there's already a way to do this!

Screen Shot 2019-05-09 at 11 32 39 AM

jpogran:
This commit introduces a new setting to allow the user to hide the
`PDK New Module` button in the Editor Title Bar.

fixes #505

### DIFF
--- a/package.json
+++ b/package.json
@@ -226,6 +226,7 @@
       "editor/title": [
         {
           "command": "extension.pdkNewModule",
+          "when": "config.puppet.titleBar.pdkNewModule.enable",
           "group": "navigation@100"
         },
         {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -423,6 +424,11 @@
             "none"
           ]
         },
+        "puppet.titleBar.pdkNewModule.enable": {
```
jpogran: Naming is hard. Tried to have a scheme that allowed us to have `puppet.<EditorComponent>.<feautre>.enable`, but open to other ideas

package.json
```diff
@@ -423,6 +424,11 @@
...
+          "type": "boolean",
+          "default": false,
```
glennsarti: Shouldn't this be `true` to keep existing functionality?

jpogran: Yes, it should. I had it default to test how it looked disabled and enabled, I missed changing it back.

jpogran:
@glennsarti default value corrected

PR #511, opened 2019-05-15
(GH-509) Update node dependencies

### LINKED ISSUE #509: Dependency update
Update the dependencies in package.json

jpogran:
This commit updates all node dependencies.

This fixes #509


jpogran:
Since this updates all packages, we need to walk through a full use of the editor because historically this has broken things

PR #521, opened 2019-05-29
(maint) Release documentation

jpogran:
This commit documentats the release process for a Marketplace releas of
the Puppet Extension.

### DIFF
--- a/tools/RELEASE.md
+++ b/tools/RELEASE.md
@@ -1,0 +1,40 @@
+# Release Puppet Extension for VSCode
+
+## Pre-release steps
+
+### Vendor Latest Puppet Language Server and Syntax Files
+
+1. Ensure [puppet-editor-services](https://github.com/lingua-pupuli/puppet-editor-services) has been tagged and release process followed. Note version.
+1. Ensure [puppet-editor-syntax](https://github.com/lingua-pupuli/puppet-editor-syntax) has been tagged and release process followed. Note version.
+1. `git clone https://github.com/lingua-pupuli/puppet-vscode` or `git clean -xfd` on working copy
+1. Update `editor-components.json` with latest `puppet-editor-services` and `puppet-editor-syntax` versions. These can be seperate commits
+1. Create PR
+
+## Prepare VSCode extension
+
+1. `git clone https://github.com/lingua-pupuli/puppet-vscode` or `git clean -xfd` on working copy
+1. Update `CHANGELOG` with all tickets from release milestone for `puppet-vscode`, and any tickets from `puppet-editor-services` and `puppet-editor-syntax` that apply to this release.
+1. Increment `version` field of `package.json`
+1. Run `npm install` to update `package-lock.json`
+1. `git commit -m '(maint) Release <version>'`
+1. Create release PR
+
+## Package VSCode extension
+
+1. `git clone https://github.com/lingua-pupuli/puppet-vscode` or `git clean -xfd` on working copy
+1. `git tag -a '<version>' -m '<version>' <commit id>`
+1. `git push <remote> <version>`
+1. `git checkout <version>`
+1. `git reset --hard <version>`
+1. `mkdir 'output'`
+1. `npm install` (this should produce no changes, but package-lock.json may be different, safe to ignore)
+1. `npx vsce package`
+1. `mv "puppet-vscode-*.vsix" 'output'`
+1. `./tools/release.ps1 -releaseversion <version> -guthubusername 'lingua-pupuli' -githubtoken <token>`
+
+## Publish VSCode extnsion
+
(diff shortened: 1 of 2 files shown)


tools/RELEASE.md
```diff
@@ -0,0 +1,40 @@
...
+1. `git reset --hard <version>`
+1. `mkdir 'output'`
+1. `npm install` (this should produce no changes, but package-lock.json may be different, safe to ignore)
+1. `npx vsce package`
+1. `mv "puppet-vscode-*.vsix" 'output'`
+1. `./tools/release.ps1 -releaseversion <version> -guthubusername 'lingua-pupuli' -githubtoken <token>`
+
+## Publish VSCode extnsion
+
+1. Install personal access token from https://pogran.visualstudio.com/puppet-vscode ([insructions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#get-a-personal-access-token))
```
glennsarti: typo `insructions`  -> `instructions`

tools/RELEASE.md
```diff
@@ -0,0 +1,40 @@
...
+1. `git clone https://github.com/lingua-pupuli/puppet-vscode` or `git clean -xfd` on working copy
+1. `npm install`
+1. `npmx vsce publish`
```
glennsarti: `npmx` ? odd, I thought it would be `npm`

tools/release.ps1
```diff
@@ -0,0 +1,168 @@
...
+  return $releaseNotes
+}
+
+$params = @{
+  VersionNumber           = $releaseVersion
+  PreRelease              = $false
+  ReleaseNotes            = (Get-ReleaseNotes -Version $releaseVersion)
+  ArtifactOutputDirectory = 'output'
+  GitHubUsername          = 'lingua-pupuli'
+  GitHubRepository        = 'puppet-editor-services'
```
glennsarti: copy-pasta fail   Should be `'puppet-editor-vscode'`

tools/release.ps1
```diff
@@ -0,0 +1,168 @@
...
+      Body        = (ConvertTo-Json $releaseData -Compress)
+    }
+    $ReleaseDetails = Invoke-RestMethod @releaseParams
+  }
+
+  # Upload assets
+  $uploadUri = $ReleaseDetails | Select -ExpandProperty upload_url
+  $uploadUri = $uploadUri -creplace '\{\?name,label\}'
+
+  Get-ChildItem -Path $artifactOutputDirectory | % {
```
glennsarti: Is this still true?

glennsarti: Should we only be attaching `.vsix` with the correct version number?

tools/RELEASE.md
```diff
@@ -0,0 +1,40 @@
+# Release Puppet Extension for VSCode
+
+## Pre-release steps
+
+### Vendor Latest Puppet Language Server and Syntax Files
+
+1. Ensure [puppet-editor-services](https://github.com/lingua-pupuli/puppet-editor-services) has been tagged and release process followed. Note version.
```
glennsarti: I think the numbering got outta whack here.

tools/release.ps1
glennsarti: Ahh I see it's part of the instructions above.  Could we instead just search the working dir for output? or change the publish npm task to do that for us?

tools/RELEASE.md
jpogran: Slight typo, thats supposed to read `npx`, not `npmx`.

>Executes <command> either from a local node_modules/.bin, or from a central cache, installing any packages needed in order for <command> to run. By default, npx will check whether <command> exists in $PATH, or in the local project binaries, and execute that. If <command> is not found, it will be installed prior to execution.

We have `vsce` in `package.json`, so it's installed by `npm install`, to use it we use `npx`

tools/RELEASE.md
jpogran: Are you are referring to the `1.` on every line, if so that's a markdown convention. Or are you are referring to the order of vendor editor services then vscode?

tools/release.ps1
jpogran: Was trying to keep source as close as possible to your original https://github.com/lingua-pupuli/puppet-editor-services/commit/048f525b57ca3629c608d0e8312ef7ea08504738, so that eventually we can use it on both with a config file or something.

`vsce publish` does not leave a `vsix` around, so it will have to be package then publish.

tools/release.ps1
glennsarti: Fixed

glennsarti:
I've added a commit to fix the release.ps1 file.

jpogran:
Rebased, squashed, ready for review and merge

PR #528, opened 2019-06-18
(GH-530) Enable Extension on Workspace files

jpogran:
This PR adds a file watcher to the `activationEvents` that initializes the extension if there are any Puppet or metadata.json files

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,9 @@
   "activationEvents": [
     "onLanguage:puppet",
     "onLanguage:puppetfile",
+    "workspaceContains:**/*.pp",
+    "workspaceContains:**/*.epp",
+    "workspaceContains:**/Puppetfile",
     "onCommand:extension.puppetShowConnectionLogs",
     "onCommand:extension.puppetShowConnectionMenu",
     "onCommand:extension.puppetLint",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how 
 
 ## [Unreleased]
 
+### Changed
+
+- ([GH-530](https://github.com/lingua-pupuli/puppet-vscode/issues/530)) Enable extension with Puppet workspace files
+
 ## [0.18.1] - 2019-06-07
 
 ### Added


package.json
```diff
@@ -53,6 +53,8 @@
   "activationEvents": [
     "onLanguage:puppet",
     "onLanguage:puppetfile",
+    "workspaceContains:**/*.pp",
+    "workspaceContains:**/metadata.json",
```
glennsarti: Curious as to the reason for the permissive glob instead of just `metadata.json`

package.json
```diff
@@ -53,6 +53,8 @@
   "activationEvents": [
     "onLanguage:puppet",
     "onLanguage:puppetfile",
+    "workspaceContains:**/*.pp",
```
glennsarti: Add epp as well?

glennsarti: Puppetfile ?

package.json
jpogran: Was thinking to catch situations like control repo, where the metadata or manifest is several levels deep

package.json
jpogran: Added

jpogran:
The current question with this PR is to leave `metadata.json` in it or not.

If `metadata.json` is not in this PR, any folder opened without Puppet manifest files won't load without opening a Puppet manifest file on purpose. Modules like the puppetlabs-powershell module do not have Puppet manifests, but do have `metadata.json` files.  If `metadata.json` is left in this PR, then any module will activate the extension.

There is risk that there is something else out there that uses a file called `metadata.json` and would cause our extension to activate when they did not intend to use it

glennsarti:
Other examples of metadata.json (Google search)

https://www.vagrantup.com/docs/boxes/format.html
https://www.chromium.org/chromium-os/build/cbuildbot-metadata-json-tutorial
https://2isa.readthedocs.io/en/latest/other/json-metadata.html
https://gitlab.gnome.org/World/ShellExtensions/desktop-icons/blob/master/metadata.json
https://fabric-shim.github.io/release-1.4/tutorial-annotated-contract-metadata.html

TBH, I'm leaning towards leave it out of this PR and raise an issue to spike trying to figure out a better way.

jpogran:
@glennsarti agreed. Amended commit and ready for merge

PR #531, opened 2019-06-19
(WIP) (maint) Use file picker for PDK New Module

jpogran:
This commit uses the builtin file picker to choose the folder to create
a new module with PDK. The previous UI required the user to type out the
fully qualified path with no autocomplete, resulting in an error prone
user experience. This presents the user with a platform native file
picker, which then passes a fully qualified path to our methods.

### DIFF
--- a/src/feature/PDKFeature.ts
+++ b/src/feature/PDKFeature.ts
@@ -1,6 +1,7 @@
 'use strict';
 
 import * as vscode from 'vscode';
+import * as path from 'path';
 import { IFeature } from '../feature';
 import { ILogger } from '../logging';
 import { PDKCommandStrings } from '../messages';
@@ -58,10 +59,14 @@ export class PDKFeature implements IFeature {
       matchOnDescription: true,
       matchOnDetail: true
     };
-    let dirOpts: vscode.QuickPickOptions = {
-      placeHolder: 'Enter a path for the new Puppet module',
-      matchOnDescription: true,
-      matchOnDetail: true
+
+    const options: vscode.OpenDialogOptions = {
+      canSelectMany: false,
+      openLabel: 'Select',
+      filters: {
+      },
+      canSelectFolders: true,
+      canSelectFiles: false,
     };
 
     vscode.window.showInputBox(nameOpts).then(moduleName => {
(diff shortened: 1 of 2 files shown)


src/feature/PDKFeature.ts
```diff
@@ -58,23 +59,33 @@ export class PDKFeature implements IFeature {
...
       matchOnDetail: true
     };
-    let dirOpts: vscode.QuickPickOptions = {
-      placeHolder: 'Enter a path for the new Puppet module',
-      matchOnDescription: true,
-      matchOnDetail: true
+
+    const options: vscode.OpenDialogOptions = {
+      canSelectMany: false,
+      openLabel: 'Select',
```
glennsarti: It would be nice if we change the title of the window as I was little lost the first time. But I can't think of a better button label than `Select`

glennsarti:
Seems to work as expected.  I did receive a message that `the term pdk ...`

image

Oh PS Core

```
C:\Source\puppet-vscode-files> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.1.0
PSEdition                      Core
GitCommitId                    6.1.0
OS                             Microsoft Windows 10.0.17763
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

src/feature/PDKFeature.ts
jpogran: I tried. It does not appear to let you change the open window title: https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions. I don't know of another way to inform the user. It's possible that this is not necessary to do with what we have planned for PDK integration

jpogran:
> Oh PS Core

Yep, Known issue. Outside scope of this project.

jpogran:
So, this PR 'fixes' the problem of picking folders for pdk new module, but increases user confusion because it doesn't inform the user what to do when it pops. We can't edit the window title or put some other message to inform the user so I'm not sure we should continue this

PR #537, opened 2019-07-15
{WIP} align hashrocket experiment

glennsarti:
The package.json lists the feature flag to be "featureFlag" with a capital F,
however the settings class expected a lowercase f.  This commit updates the
settings class to ues the expected capital F.  Without this change, no
feature flags can be set.

---

This commit adds an auto-aligner for the hash rocket (=>) for .pp files.  This
feature is hidden behind the 'hashrocket' featureflag as it may have unintended
consequences and reformat user's files unexpectedly.

The auto-align triggers on the `>` character.
* The provider then determines if the user is in a hashrocket.
* It then looks at the preceeding and proceeding lines to determine the block
  of hash rockets to consider
* Once it has block of lines, it determines the smallest hash rocket location
  and edits the text file appropriately.

### DIFF
--- a/src/feature/AlignHashRocketsFeature.ts
+++ b/src/feature/AlignHashRocketsFeature.ts
@@ -1,0 +1,125 @@
+import { IFeature } from '../feature';
+import * as vscode from "vscode";
+import { ILogger } from "../logging";
+import { IAggregateConfiguration } from "../configuration";
+
+class HashRocketLocation {
+  public rocketIndex: number;
+  public preceedingTextIndex: number;
+}
+
+class AlignHashRocketsFormatter implements vscode.OnTypeFormattingEditProvider{
+  private logger: ILogger;
+
+  constructor(logger: ILogger) {
+    this.logger = logger;
+  }
+
+  private extractHashRocketLocation(document: vscode.TextDocument, line: number): HashRocketLocation {
+    if (line < 0 || line > document.lineCount) { return null; }
+
+    var textLine = document.lineAt(line);
+    // Find the first hash-rocket `=>`
+    var hashRocket = textLine.text.indexOf(' =>');
+    if (hashRocket === -1) { return null; }
+
+    // Now find where the preceeding text ends
+    var textIndex = textLine.text.substr(0, hashRocket).trimRight().length;
+    //        key    => value 
+    //           |  | <--- this is hashRocket
+    //           | <--- this is textIndex
+    var result: HashRocketLocation = {
+      rocketIndex: hashRocket + 1,
+      preceedingTextIndex: textIndex
+    };
+
+    return result;
+  }
+
+  public provideOnTypeFormattingEdits(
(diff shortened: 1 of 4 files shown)


glennsarti:
In use

hashrocket

This formatter is a little naive and won't handle multiline hashes

genebean:
@glennsarti Does this use the logic (or code) of [Better Align](https://marketplace.visualstudio.com/items?itemName=wwm.better-align)? I ask because I have been really happy with it in my manifests.

glennsarti:
None at all.  This was an experiment.  The difference between this and BetterAlign, is this is trigger as you type.  No need for a command to format the document.

glennsarti:
To be honest this should probably use the puppet-lint lexer.

genebean:
I like the idea of it using puppet-lint

glennsarti:
We do use puppet-lint, but it doesn't do live-typing like the demo above does.  also it doesn't enforce the hashrockets the same.

src/feature/AlignHashRocketsFeature.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  public provideOnTypeFormattingEdits(
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    ch: string,
+    options: vscode.FormattingOptions,
+    token: vscode.CancellationToken):
+    Thenable<vscode.TextEdit[]>
+  {
+    //TODO: Check options for tabs vs spaces
```
jpogran: It's part of the document object IIRC, so shouldn't have to go out to options to get that

glennsarti: It was more about _inserting_ tabs etc. as right now it only inserts spaces.

src/feature/AlignHashRocketsFeature.ts
```diff
@@ -0,0 +1,125 @@
...
+  constructor(logger: ILogger) {
+    this.logger = logger;
+  }
+
+  private extractHashRocketLocation(document: vscode.TextDocument, line: number): HashRocketLocation {
+    if (line < 0 || line > document.lineCount) { return null; }
+
+    var textLine = document.lineAt(line);
+    // Find the first hash-rocket `=>`
+    var hashRocket = textLine.text.indexOf(' =>');
```
jpogran: There might not always be a space before the sign

glennsarti: there should be :-)

glennsarti:
Bouncing PR for Appveyor check

glennsarti:
Closing as won't do.  Should be done in the Language Server with proper tokenising.

genebean:
@glennsarti is there a ticket on the Language Server I should follow related to this?

glennsarti:
@genebean No ticket as yet.  When we start the work in earnest I'll migrate issue 465 to the editor-services project

PR #545, opened 2019-08-06
(GH-541) Check for latest PDK Version

### LINKED ISSUE #541: Check for latest PDK Version
**As a** user

**I would like to** have VSCode automatically check to see if my PDK install is up to date

**so that I** don't have to remember to do it myself.

jpogran:
Fixes #541

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,11 +16,13 @@ import { ConnectionHandler } from './handler';
 import { DockerConnectionHandler } from './handlers/docker';
 import { StdioConnectionHandler } from './handlers/stdio';
 import { TcpConnectionHandler } from './handlers/tcp';
-import { ConnectionType, ProtocolType, PuppetInstallType } from './settings';
+import { ConnectionType, ProtocolType, PuppetInstallType, ISettings } from './settings';
 import { ILogger } from './logging';
 import { OutputChannelLogger } from './logging/outputchannel';
 import { legacySettings, SettingsFromWorkspace } from './settings';
 import { Reporter, reporter } from './telemetry/telemetry';
+
+const axios = require('axios');
 
 export const puppetLangID = 'puppet'; // don't change this
 export const puppetFileLangID = 'puppetfile'; // don't change this
@@ -36,6 +38,7 @@ export function activate(context: vscode.ExtensionContext) {
   extContext = context;
 
   notifyOnNewExtensionVersion(extContext);
+
   checkForLegacySettings();
 
   context.subscriptions.push(new Reporter(extContext));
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -231,3 +240,63 @@ async function notifyEditorServiceDisabled(context: vscode.ExtensionContext) {
     context.globalState.update(suppressEditorServicesDisabled, true);
   }
 }
+
+async function notifyIfNewPDKVersion(context: vscode.ExtensionContext, settings:IAggregateConfiguration) {
+  const suppressPDKUpdateCheck = 'suppressPDKUpdateCheck';
+  const dontCheckAgainNotice = "Don't check again";
+  const viewPDKDownloadPage = "More info";
+
+  // if (context.globalState.get(suppressPDKUpdateCheck, false)) {
```
glennsarti: nit: can probably delete this commented code

jpogran: Actually, that was supposed to be there. That honors the user telling us to stop checking for PDK versions. Although...there is no way to ask to be notified again...

glennsarti: >  Although...there is no way to ask to be notified again

That's not a usual workflow though. "Please pester me!"

src/extension.ts
```diff
@@ -231,3 +240,63 @@ async function notifyEditorServiceDisabled(context: vscode.ExtensionContext) {
...
+async function notifyIfNewPDKVersion(context: vscode.ExtensionContext, settings:IAggregateConfiguration) {
+  const suppressPDKUpdateCheck = 'suppressPDKUpdateCheck';
+  const dontCheckAgainNotice = "Don't check again";
+  const viewPDKDownloadPage = "More info";
+
+  if (context.globalState.get(suppressPDKUpdateCheck, false)) {
+    return;
+  }
+
+  let version = '1.12.0';
```
glennsarti: I'm guessing this shouldn't be hardcoded?

glennsarti:
out-of-scope: We should probably look at linting rules for spaces around statements to be consistent

`if(settings.ruby.pdkVersion){` vs `if (settings.ruby.pdkVersion) {`

glennsarti:
Seems to work, however the message is a little odd. It asks to "click the more info button" but there is no more info button.

image

src/extension.ts
jpogran: Uncommented and verified it works on my machine :)

src/extension.ts
jpogran: Nope. Stupid mistake. fixed

jpogran:
Message was odd because I left the 'lets add popups to both conditions so I can test either way' then 'i forgot to remove the extra condition'.

glennsarti:
- [x] Confirm the more info button works as expected
- [x] Confirm message box is sane.
- [x] Confirm don't show this again works
- [x] Confirm enable/disable setting works

src/extension.ts
glennsarti: This is still commented out.  Did you forget to push your changes?

PR #548, opened 2019-08-14
Github Actions CI

jpogran:
- [x] Runs npm install, gulp, and test
- [x] Matrix of windows, linux, macos
- [x] Matrix of node versions
- [x] Increments build number
- [x] Publishes artifacts


jpogran:
So this replicates travis and appveyor. I only have it publishing the vsix in the linux platform just because there seem to be variances in paths on the different platforms.

PR #550, opened 2019-08-20
(maint) Migrate from vscode to vscode-test

jpogran:
This commit migrates the original test methods to the new vscode-test approach, as documented here: https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode


glennsarti:
- [x] can run `npm run test`
- [x] can run the extension via

PR #558, opened 2019-09-03
(GH-557) Remove Docker Connection Handler

### LINKED ISSUE #557: Deprecate and remove Docker Connection Handler
Remove docker connection handler from extension in favor of the Microsoft Remote Development pack

jpogran:
This commit removes all of the code and references for the Docker Connection Handler in favor of the Microsoft Remote Container Extension.

Fixes #557


glennsarti:
Also need to remove it from package.json

https://github.com/lingua-pupuli/puppet-vscode/blob/f94e6cb78919e74522bc5da336cfad8e62b6b6d9/package.json#L352-L360

https://github.com/lingua-pupuli/puppet-vscode/blob/f94e6cb78919e74522bc5da336cfad8e62b6b6d9/package.json#L325-L329

jpogran:
Thanks @glennsarti . Also removed from settings and added it to the legacy settings catch. I think I did that right, can you check?

src/settings.ts
```diff
@@ -116,6 +110,10 @@ export function legacySettings(): Map<string, Object> {
   let settings: Map<string, Object> = new Map<string, Object>();
   let value: Object = undefined;
 
+  // puppet.editorService.docker.imageName
```
glennsarti: The legacy stuff below is for redirecting old settings to their new location.  As this is a removal, there's no need to have this here (There's nothing to redirect to)

PR #562, opened 2019-09-05
(maint) Stale issue management

jpogran:
This uses a github-action to mark all issues older than 30 days without
activity as stale. It closes all messages not replied to within 5 days.


glennsarti:
👍 

I think we'll need to create the `awaiting-approval` and `no-issue-activity` labels prior?

glennsarti:
Our labels could do with a cleanup anyway.

jpogran:
Labels created

Issue #515, opened 2019-05-24
Integrate PDK API


jpogran:
Use PDK as a gem and not an executable.

**PDK Generate**
  - module
    - [ ] to research
  - class
    - [ ] Can use the PDK::Generate API
  - task
    - [ ] Can use the PDK::Generate API
  - transport
    - [ ] to research
  - defined type
    - [ ] Can use the PDK::Generate API

**PDK Validators**
  - metadata
    - json_lint # metadata-json-lint
      - [x] already provided by built-in json validator against published schema
    - syntax    # JSON::Pure::Parser
      - [x] already provided by built-in json validator against published schema
  - puppet
    - epp    # puppet epp validate
      - [x] already provided
    - lint   # puppet-lint
      - [x] already provided by vendored puppet-lint
    - syntax # puppet parser validate
      - [x] already provided by languageserver compile
  - ruby
    - rubocop # rubocop
      - [ ] calls rubocop gem
  - tasks
    - metadata   # JSON.parse with schema
      - [x] already provided by built-in json validator against published schema
    - name       # check if file name is valid against PDK::CLI::Util::OptionValidator.valid_task_name
      - [ ] not provided
  - yaml
    - syntax     # YAML.load
      - [ ] not provided

PR #566, opened 2019-09-16
(GH-515) Integrate PDK Ruby API

jpogran:
This commit uses the PDK Ruby API instead of the pdk shell commands to
invoke PDK commands.

- [x] PDK New Class
- [x] PDK New Defined Type
- [ ] PDK New Module
- [x] PDK New Task

Part of https://github.com/lingua-pupuli/puppet-vscode/issues/515 implementation


jpogran:
This has been shelved

PR #584, opened 2019-11-13
(maint) Update telemetry to v0.1.2

jpogran:
This commit udpates the telemetry package to 0.1.2 as well as adding the
exception method to the reporter class.

### DIFF
--- a/src/telemetry/typings/vscode-extension-telemetry.d.ts
+++ b/src/telemetry/typings/vscode-extension-telemetry.d.ts
@@ -6,6 +6,11 @@ declare module 'vscode-extension-telemetry' {
       properties?: { [key: string]: string },
       measures?: { [key: string]: number }
     ): void;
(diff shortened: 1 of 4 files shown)


src/telemetry/typings/vscode-extension-telemetry.d.ts
```diff
@@ -6,6 +6,11 @@ declare module 'vscode-extension-telemetry' {
       properties?: { [key: string]: string },
       measures?: { [key: string]: number }
     ): void;
+    sendTelemetryException(
```
glennsarti: Huh.  You woulda thought the telemetry module would have the needed typings.

jpogran: This is a result of using a wrapper class. Once we move it to it's own object we won't have to do this

PR #586, opened 2019-11-13
(GH-585) Remove deprecated settings

### LINKED ISSUE #585: Remove deprecated settings
The settings listed here have been deprecated for quite some time. It's time we should remove them:

https://github.com/lingua-pupuli/puppet-vscode/blob/master/package.json#L458-L484

jpogran:
This removes all deprecated settings. These settings have been
sending warnings for several releases now, it's time to remove them.

The code that detected the legacy settings and migrated them to the
newer settings has been kept, with commented out examples for future use

Fixes #585

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -236,14 +236,42 @@
       }
     },
     "cliui": {
-      "version": "4.1.0",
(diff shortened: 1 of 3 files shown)


jpogran:
I was going back and forth on that. I hate commenting out code, but I agree it would be worth it to keep a reference there for future legacy settings

PR #587, opened 2019-11-14
(maint) Remove TelemetryReporter Wrapper

jpogran:
This commit removes the TelemetryReporter wrapper class and type files
and global export in favor of a variable passed down through class
constructors.


jpogran:
Going to test this with real data, but otherwise I think its at the point I want it

jpogran:
Confirmed works without error in portal

PR #589, opened 2019-11-19
(GH-588) Add server telemetry pass through

### LINKED ISSUE #588: Implement telemetry event from language server
The Language Server protocol defines an [event](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#telemetry_event) that sends telemetry from the server to the client to be sent to the telemetry destination. This is defined [client side](https://github.com/microsoft/vscode-languageserver-node/blob/dc9754e7366d254e24f2eff34b23959b9cb1b0d7/client/src/client.ts#L2633-L2635) in the onTelemetry method.

jpogran:
This commit adds a listener for telemetry events from the
Language Server, and then sends them to the telemetry reporter.

Fixes #588

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -56,6 +56,10 @@ export abstract class ConnectionHandler {
       .onReady()
       .then(
         () => {
+          this.languageClient.onTelemetry(event => {
+            const eventName = event.Name ? event.Name : 'PUPPET_LANGUAGESERVER_EVENT';
+            reporter.sendTelemetryEvent(eventName, event.Measurements, event.Properties);
+          });
           this.setConnectionStatus('Loading Puppet', ConnectionStatus.Starting);
           this.queryLanguageServerStatusWithProgress();
         },


glennsarti:
Did we still want this merged for 0.22.0 ?

jpogran:
I wanted this to go in for 0.22, it's low impact as it only fires if server sends an event. It's ok by me to wait until we actually implement it server side. Otherwise this is just waiting review/merge

src/handler.ts
```diff
@@ -56,6 +56,10 @@ export abstract class ConnectionHandler {
       .onReady()
       .then(
         () => {
+          this.languageClient.onTelemetry(event => {
+            const eventName = event.Name ? event.Name : 'PUPPET_LANGUAGESERVER_EVENT';
+            reporter.sendTelemetryEvent(eventName, event.Measurements, event.Properties);
```
glennsarti: Where are the `Measurements` and `Properties` methods coming from?  I can't see them in the Language Client node module.  And the data coming from the server is `any` so there are no guarantees here.

PR #614, opened 2020-03-03
(613) Puppet Node Graph Feature Rewrite

### LINKED ISSUE #613: Port the Puppet Node Graph Visualizer to cytoscape.js
The Puppet Node Graph feature in VS Code is a useful visualization of the dependency graph between resource declarations. The viz.js library used previously for displaying the graph has not had a release in several months, the latest release changed the api and broke our usage, and the project repo has been archived in github, indicating that there isn't a future for the project.

To replace this we have chosen the cytoscape.js module, which is designed to handle scientific data-sets. This is advantageous to us, as a DAG is a common data set the library is targeted for. It also provides a large default set of visualizations, a library of extensions, and customizable theme styling which will allow us to work well within VSCode.

jpogran:
This commit replaces the viz.js Node Graph Vizualizer with a
cytoscape.js based version. All existing command ids and workflow were
kept, only the presenation logic was changed.

The cytoscape.js module is a full featured data vizualizer that allows
us to display the puppet node graph in a multitude of formats and
presentations. While this commit only replaces existing functionality,
cytoscape has alot of potential for displaying the node graph in
interesting formats since it was designed to handle very large
scientific data sets.

The latest version of the viz.js module changed it's api to a pattern
we could not use reliably in the async language server call, and the
proejct itself has been archived in github and there hasn't been a new
version in several months.

Fixes #613


package.json
```diff
@@ -527,21 +526,22 @@
   "scripts": {
     "vscode:prepublish": "pwsh -file ./build.ps1 build",
     "compile": "tsc -p ./",
-    "watch": "pwsh -file ./build.ps1 initial && tsc -watch -p ./",
+    "watch": "tsc -watch -p ./",
```
jpogran: Note to self, need to add this back in

package.json
```diff
@@ -527,21 +526,22 @@
...
     "pretest": "npm run compile",
     "test": "node ./out/test/runtest.js"
   },
   "devDependencies": {
     "@types/glob": "^7.1.1",
     "@types/mocha": "^5.2.7",
     "@types/node": "^12.0.8",
-    "@types/vscode": "^1.30.0",
+    "@types/vscode": "^1.42.0",
```
jpogran: Note to self, need to find lowest vscode version that supports webview uri

psakefile.ps1
```diff
@@ -89,6 +89,10 @@ task VendorEditorSyntax {
   Invoke-RestMethod -Uri $uri -OutFile $syntaxFilePath -ErrorAction Stop 
 }
 
+task VendorCytoscape {
+  Copy-Item -Path .\node_modules\cytoscape\dist -Recurse -Destination .\vendor\cytoscape
```
jpogran: TODO convert relative path to resolved path

package.json
jpogran: The lowest version that supports the `webview.asWebviewUri` is `1.38`. This means raising the minimum engine version to `1.38`.

glennsarti:
Is this ready for merge?

PR #628, opened 2020-04-22
(maint) Alert if jpogran.puppet-vscode installed

jpogran:
This commit adds a warning if the `jpogran.puppet-vscode` extension is
installed while the `puppet.puppet-vscode` extension is. This can cause
a race condition where each extension tries to register at the same
time, which will produce error message popups.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -35,6 +35,15 @@ let configSettings: IAggregateConfiguration;
 let extensionFeatures: IFeature[] = [];
 
 export function activate(context: vscode.ExtensionContext) {
+  let pkg = vscode.extensions.getExtension('jpogran.puppet-vscode');
+  if (pkg){
+    let message = 'The "jpogran.puppet-vscode" extension has been detected, which will conflict with the "puppet.puppet-vscode" extension. This will cause problems activating when each extension tries to load at the same time and may cause errors. Please uninstall it by executing the following from the commandline: "code --uninstall-extension jpogran.puppet-vscode"';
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -35,6 +35,15 @@ let configSettings: IAggregateConfiguration;
 let extensionFeatures: IFeature[] = [];
 
 export function activate(context: vscode.ExtensionContext) {
+  let pkg = vscode.extensions.getExtension('jpogran.puppet-vscode');
+  if (pkg){
+    let message = 'The "jpogran.puppet-vscode" extension has been detected. Please uninstall it by executing the following from the commandline: "code --uninstall-extension jpogran.puppet-vscode"';
```
glennsarti, line 40: Is it worth mentioning why this is bad?

DavidS:
```suggestion
    let message = 'The "jpogran.puppet-vscode" extension has been detected. Is not maintained anymore. To uninstall it, execute the following from the commandline: "code --uninstall-extension jpogran.puppet-vscode"';
```
?

jpogran: I think Glenn meant the kind of detail I put in the commit message could be useful to tell the user. It's less important that its unmaintained compared to the race condition that will cause errors.

DavidS: the fact that I missed it in the commit message seems to indicate that glenn is right :-D

jpogran: Updated the text

Issue #631, opened 2020-04-28
Puppet Facts View


jpogran:
**Summary of the new feature**

Add a view to the Puppet Toolbar that shows the current list of facts for the machine 

For example:

---
**As a** user

**I would like to** see a list of facts for the machine I'm developing on

**so that I** don't have to run the command myself

**and** I can select a fact to be inserted into the open manifest

Issue #630, opened 2020-04-28
Puppet Toolbar


jpogran:
**Summary of the new feature**

A new view container called Puppet Toolbar that has separate views for each type of information for different Puppet workspaces.

---
**As a** user

**I would like to** have a single window for all relevant information for the workspace I have open, whether that be a Puppet module, a Puppet Control Repo, or Bolt project.

**so that I** don't have to manually run commands to find this information seperately.

PR #632, opened 2020-04-28
GH 630 Puppet Toolbar and Puppet Facts View

jpogran:
- (GH-631) Puppet Facts View
- (GH-630) Add Puppet ToolBar

Note - Requires https://github.com/puppetlabs/puppet-editor-services/pull/243


package.json
```diff
@@ -197,8 +197,33 @@
         "command": "puppet-bolt.OpenUserInventoryFile",
         "title": "Open Bolt User Inventory File",
         "category": "Puppet"
+      },
+      {
+        "command": "puppet.refreshFacts",
+        "title": "Refresh",
+        "icon": {
+          "light": "assets/light/sync.svg",
```
jpogran, line 205: TODO: Need actual icons instead of placeholder text

src/views/facts.ts
```diff
@@ -0,0 +1,108 @@
...
+export class PuppetFactsProvider implements vscode.TreeDataProvider<PuppetFact> {
+  private elements: Array<[string, PuppetFact]> = [];
+  private _onDidChangeTreeData: vscode.EventEmitter<PuppetFact | undefined> = new vscode.EventEmitter<
+    PuppetFact | undefined
+  >();
+  readonly onDidChangeTreeData: vscode.Event<PuppetFact | undefined> = this._onDidChangeTreeData.event;
+
+  constructor(protected handler: ConnectionHandler) {
+    vscode.commands.registerCommand('puppet.refreshFacts', () => this.refresh());
+    vscode.commands.registerCommand('puppet.InsertFact', (factString) => {
```
jpogran, line 46: TODO: The problem with this is that the command is triggered on left click, not right click. So if you are scrolling and click on an element by habit, it will insert the fact wherever the cursor is in an open manifest. Need to think on the UX here

binford2k: can you register an autocomplete for values in the `$facts` hash instead?

package.json
glennsarti: There should be a nice refresh icon in the vscode list - https://github.com/microsoft/vscode-icons

src/views/facts.ts
jpogran: We have a hover provider, where you get the fact resolved when you hover over `$facts['foo']`, but don't have autocomplete yet. But! We can! It's just hard! Because puppet facts! Are! Hard!

Sorry. Might have gotten carried away there with exclamations.

After thinking this through, I am going to create an issue for inserting facts from the view and another for autocomplete for the `$facts` variable, and constrain this PR to getting the view working with the language server.

Thanks for the input! Opps, there I go again.

package.json
jpogran: Yep, this was more a reminder to go there while I got this working :)

src/views/facts.ts
glennsarti: We already autocomplete on the 1st level facts (https://github.com/puppetlabs/puppet-editor-services/blob/master/lib/puppet-languageserver/manifest/completion_provider.rb#L40-L45)

So `$facts['os']`   We don't do second or deeper level (e.g. `$facts['os']['name']`) (Because ASTs are hard)

jpogran:
@michaeltlombardi manually acceptance tested successfully. Will self merge tomorrow if @glennsarti doesn't get a chance to review

jpogran:
Manually acceptance tested successfully with new changes from @glennsarti using cached facts

Issue #647, opened 2020-05-08
Add a default welcome view to Puppet Facts View


jpogran:
Add a default welcome view to the Puppet Facts View in the Puppet Toolbar if there are no facts found, as documented [here](https://code.visualstudio.com/api/extension-guides/tree-view#welcome-content)

PR #648, opened 2020-05-08
(GH-647) Puppet Facts Welcome View

jpogran:
This commit adds a default welcome view to the Puppet Facts view if
there are no facts found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,6 +216,12 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "puppetFacts",
+        "contents": "No facts foundn\n [Refresh](command:puppet.refreshFacts)"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -216,6 +216,12 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "puppetFacts",
+        "contents": "No facts foundn\n [Refresh](command:puppet.refreshFacts)"
```
glennsarti, line 222: Does this need to be more friendlier e.g. "Please hit Refresh" or does being terse perfectly fine?

glennsarti: Nope.  Terse is fine.  Ignore.

Issue #644, opened 2020-05-08
Consolidate README


jpogran:
There is a README.md and a README_BUILD.md in the repo. This is a leftover from when there was a client and server stored in this repo and also when the readme was used as a user document. It does not need to be kept separate anymore and can be combined.

PR #653, opened 2020-05-11
(GH-644) Consolidate README

jpogran:
This commit consolidates the README and README_BUILD documents into one
README that is suitable for packaging into the extensions as well as
being a landing page for the extension in github.

It also refreshes the README content by removing most of the feature
list and table of contents sections, reodering the sections to surface
troubleshooting and installation needs first and then lists how to
develop the extension.


jpogran:
need to double check the links here in case we are still using the readme from the extension

glennsarti:
I'm still not sold on having the "clone this repo" instructions in the extension readme.  But 🤷‍♂️

Issue #663, opened 2020-05-20
server crashes when workspace path has non-standard characters


bp85:
### What Versions are you running?

OS Version: OSX 10.15.3
VSCode Version: 1.45.1
Puppet Extension Version: 0.26.1
PDK Version: 1.18.0

### What You Are Seeing?
 syntax error near unexpected token `('
[Info  - 9:30:53 PM] Connection to server got closed. Server will restart.
### What is Expected?
server will start normally.

Forgive me if I'm mistaken, but I think it's a simple fix by just adding quotes around the value of workspace.

### How Did You Get This To Happen? (Steps to Reproduce)
1. create a directory with spaces or special characters
2. add the directory to workspace
3. add some puppet code in workspace

### Output Log
```
/bin/sh: -c: line 0: syntax error near unexpected token `('
/bin/sh: -c: line 0: `/opt/puppetlabs/pdk/private/ruby/2.1.9/bin/ruby /Users/user1/.vscode/extensions/puppet.puppet-vscode-0.26.1/vendor/languageserver/puppet-languageserver --stdio --timeout=10 --local-workspace=/Users/user1 (test)/proj1--puppet-version=4.8.2'
[Info  - 9:30:53 PM] Connection to server got closed. Server will restart.
[Error - 9:30:53 PM] Connection to server got closed. Server will not be restarted.
```

glennsarti:
@jpogran Looks like a simple quoting problem.

jpogran:
Thanks @bp85 for the report, we'll get this addressed for the 0.27 release

bp85:
Appreciate the quick response @jpogran.

jpogran:
No problem @bp85 , thanks for the bug report!

PR #664, opened 2020-05-20
(GH-663) Quote Workspace path

jpogran:
This PR fixes a bug that occurs if the workspace path has a space or other non-standard character. This would cause a command line parsing error when trying to execute the language server.

### DIFF
--- a/src/helpers/commandHelper.ts
+++ b/src/helpers/commandHelper.ts
@@ -143,7 +143,8 @@ export class CommandEnvironmentHelper {
 
     args.push('--timeout=' + settings.workspace.editorService.timeout);
     if (vscode.workspace.workspaceFolders !== undefined) {
-      args.push('--local-workspace=' + vscode.workspace.workspaceFolders[0].uri.fsPath);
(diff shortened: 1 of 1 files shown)


src/helpers/commandHelper.ts
```diff
@@ -143,7 +143,8 @@ export class CommandEnvironmentHelper {
 
     args.push('--timeout=' + settings.workspace.editorService.timeout);
     if (vscode.workspace.workspaceFolders !== undefined) {
-      args.push('--local-workspace=' + vscode.workspace.workspaceFolders[0].uri.fsPath);
+      const path = `--local-workspace='${vscode.workspace.workspaceFolders[0].uri.fsPath}'`;
```
glennsarti, line 146: does the single quotes work on windows?

jpogran: Works on My Machine. Should we use double just in case?

jpogran:
Tested working on Windows 10 and macOS Catalina

Issue #666, opened 2020-05-20
Puppetfile module list in Puppet toolbar


jpogran:
**As a** user

**I would like to** list dependencies in a Puppetfile in the Puppet Toolbar

**so that I** don't have to remember to do it myself.

PR #669, opened 2020-05-23
(maint) Adds tool to update puppet-vscode projects

glennsarti:
This commit adds a helper script to synchronise (one-way) release projects in
the puppet-vscode repo.  Editor Services and Editor Syntax projects can also
be used if added to the Project description e.g.

editor-services: <a version>
editor-syntax: <a version>

Any notes etc. that are not in the correct milestones are automatically removed.

### DIFF
--- a/tools/SyncProjects.ps1
+++ b/tools/SyncProjects.ps1
@@ -1,0 +1,258 @@
+[CmdletBinding()]
+param()
+
+# Needs PowerShell 5.1+
+$ErrorActionPreference = 'Stop'
+
+$script:GithubToken = $ENV:GITHUB_TOKEN
+$script:GithubUsername = $ENV:GITHUB_USERNAME
+
+if ($null -eq $script:GithubToken) { Throw "This script requires the GITHUB_TOKEN environment variable to be set"; Exit 1 }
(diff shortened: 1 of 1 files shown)


glennsarti:
Note that this is not yet automated, this is just the script that can be used to do the automation.

jpogran:
I don't quite get what makes a card and what associates with what. I would expect the [VSCode Github Issue 671 card](https://github.com/puppetlabs/puppet-vscode/projects/3#card-39096675) to actually have issue number 666 as the title, with PR #671 and issue #666 in it. There is another [card](https://github.com/puppetlabs/puppet-vscode/projects/3#card-39096676) for #666 , with only that issue in it.

I'm not sure we can code around this, keeping in mind how issues and pull requests are the same in GIthub and how we have a very loose implementation of issues. prs, and milestones

glennsarti:
Yeah, that is one limitation. All PRs are an issue, but not all issues are a PR.

The tool uses a simple title matcher.

* Each issue (or PR) will generate a unique note (card) title
* Therefore the script can figure out if a given note title belongs to a specific issue
* Therefore because the script knows which notes belongs to which issues, it can make decisions as to whether an issue is missing  (No card), is redundant (Has card but no matching issue) or updated (Has a card for an issue but it's in the wrong column)

PR #671, opened 2020-05-28
(GH-666) Puppetfile view

jpogran:
Adds a new view to the Puppet toolbar view container that shows the puppet modules listed in the Puppetfile in the current workspace. This also adds a new command that will open the Puppetfile at the line of the selected module in the view.

**Note**: Needs an editor-services release with https://github.com/puppetlabs/puppet-editor-services/issues/256


jpogran:
0.26.1 editor-services is merged so this is clear to go

Issue #670, opened 2020-05-26
Guard installType and installDirectory


jpogran:
The extension works when there is both pdk and agent present, and when only one is present. The problem here is that the settings for the extension are not helping the user choose successful configurations and allowing mixing of settings that then produce incorrect paths to the parts of puppet we need.

The intention for `auto` was to **not** use `installDirectory` or `installType`, so that the extension would choose whether to use `pdk` or `agent` and which paths to use itself. This was intended to make the default 'happy path' the situation where a user has `pdk` installed to the default path and not require the user to set any settings.

Items to document:

-  If you installed either `pdk` or `agent` to the default install paths, and you want the extension to figure out whether to use `pdk` or `agent`, then **do not** set any settings.
-  If you installed either pdk or agent to the default install paths, but want to choose whether to use `pdk` or `agent`, set `installType` and **omit** `installDirectory`.
-  If you installed either pdk or agent to custom install paths, you **must specify** `installType` as either `agent` or `pdk`, **and set** `installDirectory` to the correct path.

_Originally posted by @jpogran in https://github.com/puppetlabs/puppet-vscode/issues/668#issuecomment-634113632_

jpogran:
https://code.visualstudio.com/api/references/contribution-points#contributes.configuration

PR #673, opened 2020-06-03
(GH-670) Guard installType and installDirectory

jpogran:
Add more explanation to the settings for installDirectory and installType using new VS Code Settings API that explains when to use each setting. Add a guard clause early on in the activation process that warns when settings are used that are incompatible.

The extension works when there is both pdk and agent present, and when only one is present. The problem here is that the settings for the extension are not helping the user choose successful configurations and allowing mixing of settings that then produce incorrect paths to the parts of puppet we need.

The intention for auto was to not use installDirectory or installType, so that the extension would choose whether to use pdk or agent and which paths to use itself. This was intended to make the default 'happy path' the situation where a user has pdk installed to the default path and not require the user to set any settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -472,17 +472,21 @@
         },
         "puppet.installDirectory": {
           "type": "string",
-          "default": null,
-          "description": "The fully qualified path to the Puppet install directory. This can be a PDK or Puppet Agent installation. For example: 'C:\\Program Files\\Puppet Labs\\Puppet' or '/opt/puppetlabs/puppet'. If this is not set the extension will attempt to detect the installation directory"
+          "markdownDescription": "The fully qualified path to the Puppet install directory. This can be a PDK or Puppet Agent installation. For example: 'C:\\Program Files\\Puppet Labs\\Puppet' or '/opt/puppetlabs/puppet'. If this is not set the extension will attempt to detect the installation directory. Do **not** use when `#puppet.installType#` is set to `auto`"
         },
         "puppet.installType": {
           "type": "string",
           "default": "auto",
-          "description": "The type of Puppet installation. Either the Puppet Development Kit (pdk) or the Puppet Agent (agent). Choose 'auto' to have the extension detect which to use automatically based on default install locations",
+          "markdownDescription": "The type of Puppet installation. Either the Puppet Development Kit (pdk) or the Puppet Agent (agent). Choose `auto` to have the extension detect which to use automatically based on default install locations",
           "enum": [
             "auto",
             "pdk",
             "agent"
+          ],
+          "enumDescriptions": [
+            "The exention will use the PDK or the Puppet Agent based on default install locations. When both are present, it will use the PDK",
+            "Use the PDK as an installation source",
+            "Use the Puppet Agent as an installation source"
           ]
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -170,6 +170,22 @@ export function SettingsFromWorkspace(): ISettings {
     pdk: workspaceConfig.get<IPDKSettings>('pdk', defaults.pdk),
   };
 
+  if (settings.installDirectory && settings.installType === PuppetInstallType.AUTO) {
+    const message =
+      "Do not use 'installDirectory' and set 'installType' to auto. The 'installDirectory' setting" +
+      ' is meant for custom installation directories that will not be discovered by the extension';
+    const title = 'Configuration Information';
+    const helpLink = 'https://puppet-vscode.github.io/docs/extension-settings';
+    vscode.window.showErrorMessage(message, { modal: false }, { title: title }).then((item) => {
```
glennsarti, line 179: Should this be an error (showErrorMessage) or a warning (showWarningMessage)

package.json
```diff
@@ -438,17 +438,22 @@
         },
         "puppet.installDirectory": {
           "type": "string",
-          "default": null,
-          "description": "The fully qualified path to the Puppet install directory. This can be a PDK or Puppet Agent installation. For example: 'C:\\Program Files\\Puppet Labs\\Puppet' or '/opt/puppetlabs/puppet'. If this is not set the extension will attempt to detect the installation directory"
+          "default": "",
+          "markdownDescription": "The fully qualified path to the Puppet install directory. This can be a PDK or Puppet Agent installation. For example: 'C:\\Program Files\\Puppet Labs\\Puppet' or '/opt/puppetlabs/puppet'. If this is not set the extension will attempt to detect the installation directory. Do **not** use when `#puppet.installType#` is set to `auto`"
```
glennsarti, line 442: Does this require a minimum version bump for vscode?  I can't remember when this was added

package.json
```diff
@@ -438,17 +438,22 @@
         },
         "puppet.installDirectory": {
           "type": "string",
-          "default": null,
-          "description": "The fully qualified path to the Puppet install directory. This can be a PDK or Puppet Agent installation. For example: 'C:\\Program Files\\Puppet Labs\\Puppet' or '/opt/puppetlabs/puppet'. If this is not set the extension will attempt to detect the installation directory"
+          "default": "",
```
glennsarti, line 441: Does this need a default?

src/settings.ts
glennsarti: From discussion - It's an error as it blows up the pathing.

jpogran:
On 1.38.1 (the only version I could get), it displays correctly, so I think we are clear to go:

image

package.json
jpogran: The `null` setting was causing a big red error when you looked at the settings in the UI, so I set it to an empty string. Testing without `default` set shows no error and allows input so I'll remove the `default` key.

package.json
jpogran: As stated in a top level comment, we are good to go on 1.38 which is our minimum

PR #680, opened 2020-06-10
(maint) Refactor Forge calls to API

jpogran:
- (maint) Extract Puppet Forge function
- (maint) Add telemetry to all features

### DIFF
--- a/src/forge.ts
+++ b/src/forge.ts
@@ -1,0 +1,105 @@
+import axios from 'axios';
+import { MarkdownString } from 'vscode';
+import { ILogger } from './logging';
+
+export interface PuppetForgeModuleInfo {
+  uri: string;
+  slug: string;
+  name: string;
+  downloads: number;
+  score: number;
+  created: Date;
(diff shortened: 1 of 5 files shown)


src/forge.ts
```diff
@@ -0,0 +1,105 @@
...
+  return new Promise((resolve) => {
+    return axios
+      .get(`https://forgeapi.puppet.com/v3/modules/${title}`, {
+        params: {
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          exclude_fields: 'readme changelog license reference',
+        },
+        headers: {
+          'Content-Type': 'application/json;charset=UTF-8',
+          'User-Agent': 'puppet-vscode/0.27.1',
```
glennsarti, line 67: Can this be programmatic?

jpogran: Yes, but I felt that was more moving parts than I could do in a refactor maint commit. Created a new issue to add that and will keep with this for this PR

Issue #691, opened 2020-06-25
Add `pdk new defined_type` command to Command Palette


greglarkin:
**As a** user

**I would like to** use a menu in the Command Palette to create a new defined type, the same way I can create a module and a class

**so that I** don't have to drop to the Terminal pane and run `pdk new defined_type ...` at the command line.

PR #692, opened 2020-06-26
(GH-691) PDK New Defined Type

jpogran:
- (GH-691) Add PDK Defined Type Command
- (maint) Fix context reference to PDK New Task
- (maint) Fix calling pdk when input is null
- (maint) Refactor pdkValidate and pdkTestUnit
- (maint) Reduce duplicatated code in pdk commands

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,6 +164,11 @@
         "title": "PDK New Task"
       },
       {
+        "command": "extension.pdkNewDefinedType",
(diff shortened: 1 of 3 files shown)


jpogran:
@glennsarti since I had to edit the class, I refactored it to how I have been wanting to do since trying the pdk ruby api approach.

Tested manually by running each command.

PR #721, opened 2021-01-14
(GH-695) Add puppet.editorService.formatOnType.maxFileSize setting

### LINKED ISSUE #695: Config setting for format on type file size limit
**Summary of the new feature**
---
**As a** user

**I would like to** be able to allow larger file sizes to be automatically formatted as I type

**so that I** don't miss this feature on large files.

glennsarti:
Fixes #695 

Now that Puppet Editor Services has a setting to change the on-type formatting
this commit updates the VSCode Extension to surface that setting to users.

- [x] Actually test this works

### DIFF
--- a/package.json
+++ b/package.json
@@ -395,6 +395,12 @@
           "default": false,
           "description": "Enable/disable the Puppet document on-type formatter, for example hashrocket alignment"
         },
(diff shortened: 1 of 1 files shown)


glennsarti:
Tested in manually.  Worked!

jpogran (approved):

PR #724, opened 2021-01-28
(GH-724) PDK New Module with Native window

jpogran:
This PR updates the `PDK New Module` command to use the native window picker instead of asking the user to enter in the fully qualified path. It also opens the newly created module inside the window the command was issued in, instead of opening a new VS Code window.

By using the native window picker, we get the correct experience on WIndows, Mac and Linxu. Even inside WSL or Remote Containers:

image

### DIFF
--- a/src/feature/PDKFeature.ts
+++ b/src/feature/PDKFeature.ts
@@ -1,5 +1,5 @@
-'use strict';
-
+import * as fs from 'fs';
+import * as path from 'path';
 import * as vscode from 'vscode';
 import { IFeature } from '../feature';
 import { ILogger } from '../logging';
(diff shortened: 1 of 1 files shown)


glennsarti (commented):
nit: typo in commit message (mmint -> maint)

src/feature/PDKFeature.ts
```diff
@@ -1,5 +1,5 @@
-'use strict';
```
glennsarti, line 1: Do we need the strict?

glennsarti (approved):

src/feature/PDKFeature.ts
jpogran: We should enable it in `tsconfig.json`, as right now its inconsistently added to the files across the project. But it probably shouldn't be removed in a feature add so I'll put it back

glennsarti:
Mergey mergey?

jpogran:
👍

Issue #730, opened 2021-03-04
Error starting extension with PDK 2.0


ShukiBT:
[puppet_server.log](https://github.com/puppetlabs/puppet-vscode/files/6082287/puppet_server.log)

### What Versions are you running?

OS Version: Win
VSCode Version : 1.53.2
Puppet Extension Version: 1.1.0
PDK Version: 2.7.2 (gem 2.7.0) 

### What You Are Seeing?
Error in VSCode when the language server for puppet starting up

### Output Log
[puppet_server.log](https://github.com/puppetlabs/puppet-vscode/files/6082384/puppet_server.log)
[puppet_server.txt](https://github.com/puppetlabs/puppet-vscode/files/6082389/puppet_server.txt)

jpogran:
I'll review the logs. In the mean time @ShukiBT, can you reply with the actual versions instead of 'latest'? It's not helpful in trying to setup a repro environment when latest can mean different things depending on when the ticket was looked at.

Also, do you mean PDK 2.0? There isn't a 2.7 out

ShukiBT:
VSCode Version : 1.53.2
Puppet Extension Version: 1.1.0
PDK Version: 2.7.2 (gem 2.7.0)

glennsarti:
Confirmed.  Using PDK 2.0.0.  Setting the appropriate ENV Vars

``` text
@ECHO OFF

SET PUPPET_BASEDIR=C:\Progra~1\Puppet~1\DevelopmentKit
SET PUPPET_NIXBASEDIR=C:/Progra~1/Puppet~1/DevelopmentKit

SET DEVKIT_BASEDIR=%PUPPET_BASEDIR%
SET RUBY_DIR=%DEVKIT_BASEDIR%\private\ruby\2.7.2
SET RUBYLIB=%PUPPET_NIXBASEDIR%/lib

SET PATH=%DEVKIT_BASEDIR%\bin;%RUBY_DIR%\bin;%PATH%
SET RUBYOPT=rubygems

SET GEM_HOME=%PUPPET_NIXBASEDIR%/share/cache/ruby/2.7.0
SET GEM_PATH=%DEVKIT_BASEDIR%\private\ruby\2.7.2\lib\ruby\gems\2.7.0;%PUPPET_NIXBASEDIR%/share/cache/ruby/2.7.0;%PUPPET_BASEDIR%\private\puppet\ruby\2.7.0
```

Running editor services from source:

``` text
[...]

PR #735, opened 2021-03-29
(GH-730) Fix ffi 1.14 libssp-0.dll missing with PDK 2.0

jpogran:
This commit adds the git binary paths to the PATH environment variable to address the missing libssp-0.dll in ffi 1.14. This can be removed once ffi 1.15 is distributed with Puppet and the PDK.

### DIFF
--- a/src/helpers/commandHelper.ts
+++ b/src/helpers/commandHelper.ts
@@ -219,6 +219,11 @@ export class CommandEnvironmentHelper {
     exe.options.env.PATH = this.buildPathArray([
       config.ruby.pdkBinDir,
       config.ruby.pdkRubyBinDir,
+      // temporary fix for ffi 1.14.2
+      // remove when pdk 2.X comes out https://github.com/puppetlabs/puppet-vscode/issues/730
+      path.join(config.ruby.puppetBaseDir, 'private', 'git', 'mingw64', 'bin'),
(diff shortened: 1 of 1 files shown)


da-ar (approved):
verified with PDK 2.0 on windows 8.1

da-ar (approved):

src/helpers/commandHelper.ts
```diff
@@ -219,6 +219,11 @@ export class CommandEnvironmentHelper {
     exe.options.env.PATH = this.buildPathArray([
       config.ruby.pdkBinDir,
       config.ruby.pdkRubyBinDir,
+      // temporary fix for ffi 1.14.2
+      // remove when pdk 2.X comes out https://github.com/puppetlabs/puppet-vscode/issues/730
+      path.join(config.ruby.puppetBaseDir, 'private', 'git', 'mingw64', 'bin'),
```
glennsarti, line 224: Do we need to guard these for Windows platforms only?  Won't stop merging though. It'll mostly be a noop on non-windows.

glennsarti (approved):

src/helpers/commandHelper.ts
jpogran: Good point, it's unnecessary for those other ones. However PDK already does this on all platforms, so 🤷 .

PR #738, opened 2021-03-30
(maint) Update codeowners with devx team

jpogran:


sanfrancrisko (approved):
I was wondering how this chicken / egg scenario would pan out

PR #754, opened 2021-06-07
(maint) Update vsce

jpogran:


jpogran:
Development dependency, so self merge

